PR #3, opened 2016-08-02
Fixed oneview-sdk version in metadata.rb

hdiomede:


metadata.rb
```diff
@@ -24,7 +24,7 @@
 chef_version     '>= 12' if respond_to?(:chef_version)
 depends          'compat_resource'
 
-gem              'oneview-sdk' if respond_to?(:gem)
+gem              'oneview-sdk', '=1.0.0' if respond_to?(:gem)
```
jsmartt, line 27: @hdiomede , I'm wondering if we need this line in there at all. In [this](https://github.com/HewlettPackard/oneview-chef/blob/master/libraries/oneview_helper.rb#L19), we already do the gem install, and as an added benefit, we can set the version based on a node attribute. If we pin a version here, it might not ever get used.

tmiotto, line 27: @jsmartt This is here because of the backward compatibility issues. The version is being fixed there so the cookbook won't get any SDK version that is not compatible with the current implementation (Notice the current cookbook implementation will not work with SDK version 2.0.0).

jsmartt, line 27: Gotcha. Ok, :+1:  from me. Do you know which resources are incompatible with 2.0.0? (We'll want to fix those up pretty quickly)

hdiomede, line 27: @jsmartt I'm working on these changes, it will be fixed as soon as possible

tmiotto:
:+1:

PR #4, opened 2016-08-04
Release 0.1.1

tmiotto:
Changes for the 0.2.0 release


jsmartt:
I like it, but am curious why 0.2.0 instead of 0.1.1 ? Pinning a version seems more like a bug fix than a minor release. I'm not even sure it matters, but that's my thought.

Also, I noticed we're getting rid of the version from the metadata. I assume this is because the helper method already installs and uses the version given as an attribute anyways, so it's not really used. Should we just get rid of the `gem` directive altogether from metadata.rb? I'm wondering if the default behavior will be to install 2.0.0, then in the helper, install and require 1.0.0

tmiotto:
Makes sense @jsmartt ,  I messed up with the release version numbers.

And... yes, it doesn't matter if we put the `gem` in metadata, it installs anyway in the `load_sdk`

hdiomede:
:+1:

PR #5, opened 2016-08-08
Fix for Ruby SDK v2.0.0

hdiomede:


hdiomede:
@jsmartt and @tmiotto what do you guys think about the cookbook version and the ruby sdk version parameter?

attributes/default.rb
```diff
@@ -12,7 +12,7 @@
 # Set which version of the SDK to install and use:
 # Warning: Changing the SDK version may cause issues within the Cookbook
 # Edit only if you know exactly what are you doing
-default['oneview']['ruby_sdk_version'] = '= 1.0.0'
+default['oneview']['ruby_sdk_version'] = '= 2.0.0'
```
tmiotto: I suggest fixing version to `"~> 2.0"`

jsmartt:
@hdiomede , I think the cookbook version is good, but for the sdk version, we can do `~> 2.0` (or `~> 2.0.0` at the least). We won't want to have to update this cookbook for bug fixes in the SDK, or minor version improvements either for that matter.

libraries/oneview_resource_base.rb
```diff
@@ -36,54 +38,77 @@ def create_or_update(item = nil)
           Chef::Log.info("#{resource_name} '#{name}' is up to date")
         else
           Chef::Log.debug "#{resource_name} '#{name}' Chef resource differs from OneView resource."
-          Chef::Log.info "Update #{resource_name} '#{name}'"
-          converge_by "Update #{resource_name} '#{name}'" do
-            item.update(temp) # Note: Assumes resources supports #update
+          Chef::Log.info "#{method_2.to_s.capitalize} #{resource_name} '#{name}'"
+          converge_by "#{method_2.to_s.capitalize} #{resource_name} '#{name}'" do
+            item.update(temp)
```
jsmartt, line 43: Also, did you mean to do a `item.send(method_2)` here?

hdiomede, line 43: we don't have an edit method....update was kept the same. It was just something to fix the message

metadata.rb
```diff
@@ -16,7 +16,7 @@
 description      'Provides OneView resources'
 long_description 'Provides resources for configuring HPE OneView-managed devices'
 
-version          '0.1.1'
+version          '1.0.0'
```
tmiotto: Only some internal methods are edited, so maybe `0.1.2`?

tmiotto:
Regarding the Cookbook version, the plan is not to release yet, but it would be good to change the version so it can be different from the previous release.

My suggestion is to change it to the next master version and add one suffix like `-dev` or `-alpha`, since we don't know if our next release would be a _bugfix_, _minor update_ or _major update_.

Following that, the version here would be `1.0.0-dev`, but I'm open to suggestions. Any other ideas @jsmartt  @hdiomede ?

jsmartt:
I'd say if we're not going to release, we don't need to mess with the version at all. I'd also avoid the -dev or -alpha versions and just keep it simple. When we're ready to release, we can bump the version.

PR #6, opened 2016-08-10
Feature/enclosure group

hdiomede:


libraries/matchers.rb
```diff
@@ -16,7 +16,7 @@
   oneview_resources = {
     oneview_resource:                   standard_actions,
     oneview_enclosure:                  [:add, :remove],
-    oneview_enclosure_group:            standard_actions,
+    oneview_enclosure_group:            [:create, :create_if_missing, :delete, :set_script],
```
tmiotto: I think it would be better to describe it as: `standard_actions + [:set_script]`

CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 ### New
   - Upgraded Ruby SDK version to 2.0.0
   - Fixed add/create, delete/remove for resources
+  - Enclosure group actions
```
tmiotto: "_Added support to Enclosure group actions_" seems much more like a change instead just saying "_Enclosure group actions_"

PR #7, opened 2016-08-11
Feature: Ethernet Network

tmiotto:
Added support to `:bulk_create`

### DIFF
--- a/spec/fixtures/cookbooks/oneview_test/metadata.rb
+++ b/spec/fixtures/cookbooks/oneview_test/metadata.rb
@@ -2,6 +2,6 @@ name             'oneview_test'
 maintainer       'test cookbook'
 license          'All rights reserved'
 description      'A test cookbook for the oneview cookbook'
-version          '0.1.0'
+version          '0.1.1'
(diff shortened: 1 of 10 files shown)


jsmartt:
CHANGELOG?

spec/fixtures/cookbooks/oneview_test/metadata.rb
```diff
@@ -2,6 +2,6 @@
 maintainer       'test cookbook'
 license          'All rights reserved'
 description      'A test cookbook for the oneview cookbook'
-version          '0.1.0'
+version          '0.1.1'
```
hdiomede, line 5: Do we need to change the version for this cookbook. It's just for testing!

hdiomede:
:+1:

spec/fixtures/cookbooks/oneview_test/metadata.rb
tmiotto, line 5: No, we don't. It was just to keep in track with the original cookbook version...

PR #11, opened 2016-08-15
Enclosure full support implementation

hdiomede:


resources/enclosure.rb
```diff
@@ -32,3 +34,20 @@
...
   remove
 end
+
+action :reconfigure do
+  item = load_resource
+  item.retrieve!
+  item.configuration
+end
+
+action :refresh do
```
jsmartt, line 51: For these actions that don't test anything before taking an action, it might be a little misleading to people running the client with the --whyrun option. We say it's supported for the whole resource [here](https://github.com/HewlettPackard/oneview-chef/blob/master/libraries/oneview_resource_base.rb#L114), but these 2 new actions don't actually support it.

Is there anything we can add here as a test? Ie only do a reconfigure if it's in an inconsistent state or something?

tmiotto, line 51: For `:reconfigure` action it is possible to check the `enclosure['state']` before and do it only if it is `inconsistent`.

But for the `:refresh` action there is no flag indicating it needs to be refreshed. We cannot check if there are changes to do the refresh because the refresh is the action that does it itself.

hdiomede, line 51: A good way to trigger a refresh would be using an attribute when calling a CHEF action that might need a refresh after the execution.
That can be done by the notify.

Regarding the reconfigure I agree with Thiago!

jsmartt, line 51: Cool. Yeah, sounds good. Agreed that notify is definitely the way to use these :)

tmiotto:
:+1:

PR #12, opened 2016-08-17
Network set implementation

tmiotto:
Added support to Network set actions: `:create` , `:create_if_missing`, `:delete`


hdiomede:
:+1:

PR #14, opened 2016-08-24
Datacenter implementation

hdiomede:


jsmartt:
Looks good :smile: Can you add a section to the README though with the usage of this resource?

hdiomede:
Of course, I forgot we were doing it.
Thanks!

jsmartt:
Cool. And it's not necessarily specific to this PR/resource, but we have a few different places that examples live (README, the examples/ directory, and to an extent, the spec/fixtures/cookbooks/oneview_test/recipes directory).  I think it's fine; I really like having a few examples in the README, but putting a complete set of them in there would get crazy (hence the examples/ directory). I think we just need to put a link in the README to that and the test cookbook examples (and keep a pretty complete, well-commented set in the examples directory).

Just curious if you guys had any thoughts about it?

hdiomede:
@jsmartt I've just created #15 to address this question

PR #16, opened 2016-08-31
Connection template

tmiotto:
Added support to connection template:
- By individual resource (HPE OneView Rest API approach)
- Integrated with Ethernet network (More user-friendly:  HPE OneView GUI approach)


hdiomede:
:+1:

PR #20, opened 2016-09-02
Fix storage_system action type in README

jsmartt:
Should be :remove, not :delete.

Matches https://github.com/HewlettPackard/oneview-chef/blob/master/resources/storage_system.rb#L27


hdiomede:
:+1:

PR #22, opened 2016-09-02
Polished volume resource

### LINKED ISSUE #21: Volume resource syntax bug
https://github.com/HewlettPackard/oneview-chef/blob/master/resources/volume.rb#L51 is missing an `&`. There's a single `&` that should be `&&`. Currently throws an error similar to:

```
    ================================================================================
    Error executing action `create` on resource 'oneview_volume[CHEF_VOL_01]'
    ================================================================================

    NoMethodError
    -------------
    undefined method `&' for "ThreePAR7200-3167":String

    Cookbook Trace:
    ---------------
    /home/vagrant/chef-repo/.chef/local-mode-cache/cache/cookbooks/oneview/resources/volume.rb:51:in `load_storage_system'

```

Easy fix, but we'll want to add some tests too.

jsmartt:
Fixes #21 

Also, it enables you to pass the same data attributes that you would for an update, but for a create. You no longer need the `provisioningParameters`; see the new examples for more of a description on this.

Also, it:
- Adds test for the volume resource. Somehow those were missing...
- Makes a small fix in the oneview_helper library (load_resource) that converts to and from JSON. This fixes issues I was seeing for data with nested hashes containing symbols. The resource `:like` method was reporting they were different when the only difference was the type was a symbol instead of a string.
- Makes a small fix in the base resource to do a real, full clone of the data instead of a shallow copy. For nested data hashes, it would override the copy when we did a retrieve.

### DIFF
--- a/libraries/oneview_resource_base.rb
+++ b/libraries/oneview_resource_base.rb
@@ -31,7 +31,7 @@ module OneviewCookbook
     def create_or_update(item = nil, method_1 = :create, method_2 = :update)
       ret_val = false
       item ||= load_resource
-      temp = item.data.clone
+      temp = Marshal.load(Marshal.dump(item.data))
       if item.exists?
         item.retrieve!
         if item.like? temp
--- a/resources/volume.rb
+++ b/resources/volume.rb
@@ -27,16 +27,30 @@ action_class do
   # @return [OneviewSDK::Volume] Loaded Volume resource
   def load_resource_with_associated_resources
     item = load_resource
-    item = load_storage_system(item)
+    load_storage_system(item)
 
     # item.set_storage_pool(OneviewSDK::StoragePool.new(item.client, name: storage_pool)) if storage_pool
-    # HACK Ruby SDK 1.0.0 issue workaround
-    sp = OneviewSDK::StoragePool.new(item.client, name: storage_pool)
-    raise "Storage Pool '#{sp['name']}' not found" unless sp.retrieve!
-    item['provisioningParameters']['storagePoolUri'] = sp['uri']
+    # Workaround for issue in oneview-sdk:
+    if storage_pool
+      sp = OneviewSDK::StoragePool.find_by(item.client, name: storage_pool).first
+      raise "Storage Pool '#{sp['name']}' not found" unless sp
+      item['storagePoolUri'] = sp['uri']
+    end
 
     item.set_snapshot_pool(OneviewSDK::StoragePool.new(item.client, name: snapshot_pool)) if snapshot_pool
-    item.set_volume_template(OneviewSDK::VolumeTemplate.new(item.client, name: volume_template)) if volume_template
+    item.set_storage_volume_template(OneviewSDK::VolumeTemplate.new(item.client, name: volume_template)) if volume_template
+
+    # Convert capacity integers to strings
+    item['provisionedCapacity'] = item['provisionedCapacity'].to_s if item['provisionedCapacity']
+    item['allocatedCapacity'] = item['allocatedCapacity'].to_s if item['allocatedCapacity']
+
+    unless item.exists? # Also set provisioningParameters if the volume does not exist
+      item['provisioningParameters'] ||= {}
+      item['provisioningParameters']['shareable'] = item['shareable'] if item['provisioningParameters']['shareable'].nil?
+      item['provisioningParameters']['provisionType'] ||= item['provisionType']
+      item['provisioningParameters']['requestedCapacity'] ||= item['provisionedCapacity']
+      item['provisioningParameters']['storagePoolUri'] ||= item['storagePoolUri']
+    end
     item
   end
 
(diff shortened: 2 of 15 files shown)


tmiotto:
@jsmartt  I agree with this PR, but I think we still need to update the documentation (README) accordingly. Right now, if I look in the OneView REST API and in the README of this project, I cannot use the resource properly because I don't have the right information from each of them.

jsmartt:
@tmiotto , I'm not sure I follow. The usage is actually the same, but I've just added a few additional intelligent features to make things easier. I tried to create a very detailed example in the examples directory, with lots of comments.

There isn't an example in the README, but we don't really have full examples for hardly any resource in the README; just usage info, describing which options and actions are available.

tmiotto:
I was mentioning that if the user read the README of the Chef and the README of the SDK he would have no clue on how to use the new way you created.

I mean, I would never know if wasn't for the example that I don't need to use the `:provisioningParameters` hash, and also the `provisionedCapacity` parameter is not documented elsewhere. 

Since it's different from the default usage I think it would be good to have at least a "You do not need to decalre your provisioning parameters inside the `provisioningParameters` hash" and something like "The :requestedCapacity can be also used as :provisionedCapacity" if you don't want to list the new _parameter aliases_ added.

libraries/oneview_resource_base.rb
```diff
@@ -31,7 +31,7 @@ module ResourceBase
     def create_or_update(item = nil, method_1 = :create, method_2 = :update)
       ret_val = false
       item ||= load_resource
-      temp = item.data.clone
+      temp = Marshal.load(Marshal.dump(item.data))
```
tmiotto, line 34: Why this?

resources/volume.rb
```diff
@@ -27,16 +27,30 @@
...
+      item['storagePoolUri'] = sp['uri']
+    end
 
     item.set_snapshot_pool(OneviewSDK::StoragePool.new(item.client, name: snapshot_pool)) if snapshot_pool
-    item.set_volume_template(OneviewSDK::VolumeTemplate.new(item.client, name: volume_template)) if volume_template
+    item.set_storage_volume_template(OneviewSDK::VolumeTemplate.new(item.client, name: volume_template)) if volume_template
+
+    # Convert capacity integers to strings
+    item['provisionedCapacity'] = item['provisionedCapacity'].to_s if item['provisionedCapacity']
+    item['allocatedCapacity'] = item['allocatedCapacity'].to_s if item['allocatedCapacity']
```
tmiotto, line 45: Where this `allocatedCapacity` is coming from? It is another alias for the `requestedCapacity` ? If so, it shouldn't be used in line 51?

resources/volume.rb
```diff
@@ -27,16 +27,30 @@
...
+
+    unless item.exists? # Also set provisioningParameters if the volume does not exist
+      item['provisioningParameters'] ||= {}
+      item['provisioningParameters']['shareable'] = item['shareable'] if item['provisioningParameters']['shareable'].nil?
+      item['provisioningParameters']['provisionType'] ||= item['provisionType']
+      item['provisioningParameters']['requestedCapacity'] ||= item['provisionedCapacity']
```
tmiotto, line 51: Why I can use just `item['provisionedCapacity']` but cannot use simply `item['requestedCapacity']` too?

libraries/oneview_resource_base.rb
jsmartt: @tmiotto , for nested data hashes, it would override the copy when we did a retrieve. The clone method is just a shallow copy, so if you play with how it works on nested hashes, second-level hashes or greater are just pointers to the original. The retrieve will overwrite that copy, causing the like? method to say they're the same, when they might not be.

resources/volume.rb
jsmartt: You shouldn't have to set it, but if a user would happen to set it as an integer instead of a string, it would cause the comparison to report that they're different. This just converts it if they set it. If not, it does nothing. (Makes it a little more fool-proof :smile: )

resources/volume.rb
jsmartt: This section is just allowing you to set the capacity like you would if you did an update. `item['provisionedCapacity']` exists once the volume is created, but `item['requestedCapacity']` does not (it only exists in the create, and it's `item['provisioningParameters']['requestedCapacity']`.

Although the API requires slightly different data for creation than updates, Chef should be as declarative as possible. In my recipe I define it as it will exist after it's created, and let our resource figure out how to create it if it does not exist.

Hopefully that makes sense. If not, I can give an example...

tmiotto:
Got it! It was confuse because of OneView implementation...

I would just add one line recommending to use the `update` synthax since calling an update with the `provisioningParameters` hash may cause some disruption in the recipes.

resources/volume.rb
```diff
@@ -27,16 +27,30 @@
...
-    item.set_volume_template(OneviewSDK::VolumeTemplate.new(item.client, name: volume_template)) if volume_template
+    item.set_storage_volume_template(OneviewSDK::VolumeTemplate.new(item.client, name: volume_template)) if volume_template
+
+    # Convert capacity integers to strings
+    item['provisionedCapacity'] = item['provisionedCapacity'].to_s if item['provisionedCapacity']
+    item['allocatedCapacity'] = item['allocatedCapacity'].to_s if item['allocatedCapacity']
+
+    unless item.exists? # Also set provisioningParameters if the volume does not exist
+      item['provisioningParameters'] ||= {}
+      item['provisioningParameters']['shareable'] = item['shareable'] if item['provisioningParameters']['shareable'].nil?
```
tmiotto, line 49: I'm curious, why this is not like the others? (`item[x][y] ||= item[y]`)

jsmartt: It's effectively a `||=`, but since the value can be false, the ||= will override false values **no matter what** (even if the override is nil). To demonstrate...

``` ruby
shareable = false
 => false
shareable ||= nil
 => nil
```

tmiotto:
Good to go :+1:

PR #24, opened 2016-09-02
Fix Travis issues

### LINKED ISSUE #18: Travis job for Ruby 2.1 is failing when is trying to install buff-extensions (2.0.0)
Travis job for Ruby 2.1 is failing and outputting the following message:

> An error occurred while installing buff-extensions (2.0.0), and Bundler cannot continue.
> Make sure that `gem install buff-extensions -v '2.0.0'` succeeds before bundling.

jsmartt:
Fixes #18

Instead of locking versions in the Gemfile for gems that are unimportant, this just removes Ruby 2.1, replacing it with 2.3.1. Chef ships with newer versions of Ruby than 2.1 anyways.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,9 @@
 language: ruby
 rvm:
-  - 2.1
   - 2.2.5
+  - 2.3.1
+
+before_install: gem update bundler
 
 script:
   - rake test
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,6 @@
 language: ruby
 rvm:
+  - 2.1
   - 2.2.5
   - 2.3.1
 
--- a/Gemfile
+++ b/Gemfile
@@ -1,8 +1,12 @@
 source 'https://rubygems.org'
 
-gem 'chef', '>=12.0.0'
+gem 'chef', '>= 12.0.0'
 gem 'chefspec'
-gem 'berkshelf'
+if RUBY_VERSION =~ /^2\.[01]/
+  gem 'berkshelf', '~> 4.3' # Use older version of berkshelf for Ruby 2.0 and 2.1
+else
+  gem 'berkshelf'
+end
 gem 'foodcritic'
 gem 'rubocop', '= 0.40.0'
 gem 'oneview-sdk'
--- a/attributes/default.rb
+++ b/attributes/default.rb
@@ -12,7 +12,7 @@
 # Set which version of the SDK to install and use:
 # Warning: Changing the SDK version may cause issues within the Cookbook
 # Edit only if you know exactly what are you doing
-default['oneview']['ruby_sdk_version'] = '~> 2.0'
+default['oneview']['ruby_sdk_version'] = '~> 2.1'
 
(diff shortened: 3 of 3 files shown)


jsmartt:
Hmm. That's interesting. Now it's failing [here](https://github.com/HewlettPackard/oneview-sdk-ruby/blob/master/lib/oneview-sdk/resource.rb#L340) in the SDK. Maybe Ruby 2.3 changed how ObjectSpace works...?

jsmartt:
[SDK PR 86](https://github.com/HewlettPackard/oneview-sdk-ruby/pull/86) should fix these new issues.

jsmartt:
OK, @hdiomede & @tmiotto , now the 2.3.1 build is passing after the SDK release.

Note that I saw some issues on 2.3.1 with older versions of the compat_resource cookbook (~12.10), so if you run into anything, make sure that's up to date. Here's what I got before updating to 12.14.0: 

```
NoMethodError:
  undefined method `full_gem_path=' for #<Gem::Specification:0x3260a18 ->
  Did you mean?  full_gem_path
# /tmp/d20160906-1050-i0nsp/cookbooks/compat_resource/files/lib/compat_resource/gemspec.rb:10:in `block in <module
# /tmp/d20160906-1050-i0nsp/cookbooks/compat_resource/files/lib/compat_resource/gemspec.rb:8:in `new'
...
```

tmiotto:
Since we are having problems with the compat_resource ~12.10, I think we can bump up the compatible version to `>=12.14`  for the `depends "compat_resource"` in metadata.rb.

What do you think @jsmartt ?

tmiotto:
Other ideas that just came in my mind right now is specifying it as `"~> 12"` and restricting with a `conflicts "compat_resource", "< 12.14"`, or simply locking it as `"~> 12.14"`.

jsmartt:
I kinda like `~> 12.14`, if anything. That way it won't pull down any breaking changes, should a 13.0 be released. If it continues to cause issues, I'd vote to just remove it as a dependency altogether; I'm not sure it's really providing us much value (just backwards compatibility with a 1+ yr old chef client, even though we depend on a brand new version of compat_resource).

jsmartt:
Keep in mind that this is also only necessary for Ruby 2.3.1, which doesn't even come with the latest version of Chef. I just installed the latest Chef version on a CentOS 7.2 box, and it comes with Ruby version 2.1.9. I'd just say to leave the dependencies as is

jsmartt:
Also, what if we kept 2.1 in our Travis tests and just do something like this in our gemfile?:

``` ruby
if RUBY_VERSION =~ /^2\.[01]/
  gem 'berkshelf', '~> 4.3' # Use older version of berkshelf for Ruby 2.0 and 2.1
else
  gem 'berkshelf'
end
```

Since Chef uses Ruby 2.1, I think we best keep it in the tests.

jsmartt:
Cool. Should be good to go now unless there are any more thoughts?

tmiotto:
Alright :+1:

PR #26, opened 2016-09-06
Added diff output for updating resources

### LINKED ISSUE #17: Show diff of resource that needs updated
When updating resources, there is no log output that indicates what is changed; just that it needs updated. However, when trying to debug a resource that continually needs updated because it seems to never match the definition, it's impossible to tell why.

For instance, if you have:

``` ruby
oneview_ethernet_network 'Chef-Eth-Net' do
  client <my_client>
  data(
    vlanId: 51,
    purpose: 'General',
    smartLink: true,
    privateNetwork: false
  )
end
```

It will always get "updated". On further investigation, it's because `"smartLink"=>false` on OneView, and it can't be set to false. It's pretty easy to output a diff for single-level hashes like ethernet_network data. You could just output:

``` bash
...
[2016-09-01T19:48:39+00:00] INFO: Update oneview_ethernet_network 'Chef-Eth-Net'

    - Update oneview_ethernet_network 'Chef-Eth-Net'

  key1: old_val -> new_val
  key2: old_val -> new_val
...
```

However, for nested hashes, this may prove a little more difficult. We could do a dot chain, which could look something like:

``` bash
...
[2016-09-01T19:48:39+00:00] INFO: Update oneview_ethernet_network 'Chef-Eth-Net'

    - Update oneview_ethernet_network 'Chef-Eth-Net'

  key1: old_val -> new_val
  first_level.key2: old_val -> new_val
  first_level.second.third.key3: old_val -> new_val
...
```

The easiest option would just be to print a debug statement with both complete hashes (no diff), but I think it's worth handling cleaner than that. Thoughts?

jsmartt:
### Description

This PR applies to resource updates. It:
- Prints resource diffs in a simplified format:
  
  ```
  [2016-09-19T16:24:44+00:00] INFO: Update oneview_resource 'ServerProfile1'
  
    - Would Update oneview_resource 'ServerProfile1'
  
      description: Test Server Profile -> Test Server Profile created by Chef
      boot:
        manageBoot: false -> true
        order: [] -> ["PXE", "CD", "Floppy", "USB", "HardDisk"]
  ```
  - As you can see, it handles top-level attributes (e.g. description), as well as nested hashes, arrays, booleans, etc. (handles type mismatches also)
- Also prints a debug log statement of the current and desired states of the resource

It should be safe for all resources to use; although there may be some unexpected formatting, it rescues StandardErrors, prints an error to the logger, and returns an empty diff string.
### Issues Resolved

Fixes #17 
### Check List


libraries/oneview_helper.rb
```diff
@@ -102,5 +102,44 @@ def convert_keys(info, conversion_method)
       end
       support
     end
+
+    # Get the diff of the current resource state and the desired state
+    # @param [OneviewSDK::Resource] resource Resource containing current state
+    # @param [Hash] desired_data Desired state for the resource
+    # @return [String] Diff string (multi-line)
```
tmiotto: I cannot see the expected behavior for _no diff_ or _error_ in the documentation comments.

Are those empty strings, nil...?

jsmartt: It returns an empty string if there is no diff. Also, the rescue below handles exceptions and just returns an empty string.

tmiotto: Yeah, I got it looking at the code... but it would be good to have in the comments if we are planning to generate documentation.

jsmartt: @tmiotto , the yard docs are updated now

tmiotto:
Is this ok to be reviewed?

jsmartt:
Yep! (I think so anyways) :smile:

tmiotto (approved):

PR #27, opened 2016-09-09
Logical Interconnect Group

tmiotto:
- Added idempotence support to actions
- Added tests
- Polished examples

:memo: _Note: These changes are not needed to be added into the README or CHANGELOG_

### DIFF
--- a/examples/logical_interconnect_group.rb
+++ b/examples/logical_interconnect_group.rb
@@ -14,52 +14,59 @@
 # FC Network: FCNetwork1
 
 
-client = {
+my_client = {
   url: '',
   user: '',
   password: ''
 }
 
+# LOGICAL INTERCONNECT GROUP 1 #
 # Simple Logical Interconnect group creation without interconnects and uplinks
-lig_data = {
-  enclosureType: 'C7000',
-}
-
 oneview_logical_interconnect_group 'LogicalInterconnectGroup1' do
-  client client
+  client my_client
   data(enclosureType: 'C7000')
-  action :create
 end
+################################
 
+
+# LOGICAL INTERCONNECT GROUP 2 #
 # Logical Interconnect group creation with interconnects but without uplinks
 oneview_logical_interconnect_group 'LogicalInterconnectGroup2' do
-  client client
+  client my_client
   data(enclosureType: 'C7000')
+  # Define each interconnect type in the corresponding bay
+  # If not specified the interconnect is not added to the group
   interconnects [
     {bay: 1, type: 'HP VC FlexFabric 10Gb/24-Port Module'},
     {bay: 2, type: 'HP VC FlexFabric 10Gb/24-Port Module'}
   ]
-  action :create
 end
+################################
 
 
-# Complete Logical Interconnect group creation with interconnects and uplinks
+# LOGICAL INTERCONNECT GROUP 3 #
+# Complete Logical Interconnect group creation with interconnects and uplinks.
+# We are specifying first the Uplink set data outside the chef block to make it more clear.
 
 # Ethernet uplink set
+## First specify the main attributes
 lig_03_uplink_01_data = {
   name: 'LogicalInterconnectGroup3 - UplinkSet1',
   networkType: 'Ethernet',
   ethernetNetworkType: 'Tagged'
 }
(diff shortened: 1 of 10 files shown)


resources/logical_interconnect_group.rb
```diff
@@ -38,15 +40,57 @@ def load_lig(item)
...
       end
-      uplink_info[:connections].each { |link| up.add_uplink(link[:bay], link[:port]) }
+      parsed_uplink_info[:connections].each { |link| up.add_uplink(link[:bay], link[:port]) }
       item.add_uplink_set(up)
     end
     item
   end
+
+  # like? function specific to the interconnectMapTemplate resource
+  def interconnect_map_like?(item, another_item)
```
jsmartt: Would this be a good thing to just override in the SDK? (That way we won't have to have this sort of custom logic split out into different providers like this, and we can still use the generic create_or_update method on this and other resources)

examples/logical_interconnect_group.rb
```diff
@@ -23,19 +23,18 @@
 # Simple Logical Interconnect group creation without interconnects and uplinks
 lig_data = {
   enclosureType: 'C7000',
-  type: 'logical-interconnect-groupV3'
 }
```
jsmartt: I like this example better than before :) There's just a few things I noticed or would suggest:
- It doesn't look like the `lig_data` object is ever used. It probably just got left in on accident.
- On many of our examples we name our fake client object "client". This may cause some confusion for people when they see the resource that contains `client client`. I'm going to try to name the client object `my_client` in the future to avoid this confusion. I'd suggest it here too.
- Lastly, for the resources where we're using the default action, I'd recommend leaving it out altogether (default is :create for this resource). When we leave it out, we'd obviously want to have a comment in there somewhere saying that it's the default, and that's why we left it out. This one is not very important because it works with or without it, but it may simplify some of the examples.

And in general, the more comments in these examples, the better :smile:

resources/logical_interconnect_group.rb
tmiotto: Yeah... It would probably be the better alternative, it would simplify a lot the work on the Chef resources.

Do you think we should remove these things and leave the comparison to the Ruby SDK, and if it's not working as we expect we should open an issue there? @jsmartt @hdiomede

hdiomede: Moving this logic to the Ruby SDK would be a good idea or maybe implement a generic method that deals with two hashes, arrays and all kinds of types.

tmiotto: It is not the case, there is some attributes that OneView adds to the LIG after the creation, so just creating method that deals with hash or arrays won't do the trick, it needs a specific filter to classify what are the configurations that need to be compared.

I don't know if we could add anything else in the general `like?` method right now. For example, to deal with arrays inside the LIG I don't care about how it is ordered (so I `sort`them), but inside another resource the order may be important and sorting may break the comparison, so these kind of changes may work for some resources but not for others since we don't have things like this mapped.

jsmartt: Yeah, I think we could definitely come up with something good in the SDK and just let the Chef cookbook struggle with the updates part of it for a while.
In the SDK, we can just override the like? or recursive_like? method to handle that specific resource correctly (the same way we do create, etc.).

tmiotto: Yep, that's my point, we could not add anything else to the resource.rb `like?` because we don't know if these things are general, and because for each resource there may be some particularities, so overriding the resources would be the way to handle this.

tmiotto:
So now we have two options:
1.  Leave it as is until we have the `like?` overriden in the SDK
2.  Remove it and treat it as an _issue_ in the SDK

The first option leaves the code undone since we will still need to remove the comparisons in the future, and the second option leaves the code done since it will be the _"final"_ version of it.

@jsmartt @hdiomede  What do you think?

jsmartt:
I'd vote for less work in the future. Remove it and open an SDK issue. This is partly because I'm not confident in my ability to remember to remove it later on :confused: 

Also, this is a bigger issue than just this resource. I think there are a few that need to do an override.

hdiomede:
:+1:

Issue #31, opened 2016-09-14
Implementation of Server Hardware


hdiomede:
Implement the Server Hardware resource with its actions and unit tests.

Actions:
- `add`
- `add_if_missing`
- `remove`
- `refresh`
- `set_power_state`
- `update_ilo_firmware`

PR #32, opened 2016-09-14
Server hardware implementation

hdiomede:
### Description

Implementation for the Server hardware resource
### Issues Resolved
- #31 
### Check List


examples/server_hardware.rb
```diff
@@ -0,0 +1,48 @@
...
+
+oneview_server_hardware '172.18.6.4' do
+  data(
+    hostname: '172.18.6.4',
+    username: 'user',
+    password: 'password',
+    licensingIntent: 'OneViewStandard',
+    configurationState: "Monitored"
+  )
+  client my_client
```
tmiotto: put the `client` on top to keep the pattern

jsmartt:
:+1:  I'm good with this PR, but I think there is 1 thing that needs to be changed in the SDK: being able to retrieve and check for existence based on the hostname, not just name and URI. If you set the name of a server_hardware resource to something other than the hostname, it won't be able to retrieve it, so on every run, it'll try to re-add it (resulting in an error). Again, I think this is an SDK issue, not an issue with this cookbook (see [#91](https://github.com/HewlettPackard/oneview-sdk-ruby/issues/91)).

jsmartt:
Also, the Ruby 2.1 build is unhappy again because of another gem requiring Ruby 2.2. Not our problem, so I don't think we need to hold off on any merging because of it.

PR #38, opened 2016-09-15
Added generic custom matcher builder

tmiotto:
### Description

Adds automatic generation for custom matchers without the need to define them manually.
### Check List
- [ x ] New functionality includes testing.
  - [ x ] All tests pass (`$ rake test`).
  - [ x ] New resources and/or actions have are included in the [matchers.rb](https://github.com/HewlettPackard/oneview-chef/blob/master/libraries/matchers.rb) and [matchers_spec](https://github.com/HewlettPackard/oneview-chef/blob/master/spec/unit/resources/matchers_spec.rb).


jsmartt:
Nice! So much better than what we had before :smile: I'll be good with this once we get the Travis issues fixed to verify it works on Ruby 2.1

PR #40, opened 2016-09-23
Make #load_sdk do the right thing

jsmartt:
### Description

This PR make sure that the correct version of the SDK is loaded when possible, and gives you helpful log info if it can't be.
### Issues Resolved

No GH issues, but before, the require in the rescue block would load the latest version of the SDK, not necessarily the one you just installed. You can demonstrate this behavior by setting the version attribute to an older version than is currently installed, and it will install the older version, but actually load the newest one.

Now, it will try to load the correct SDK version, and if it fails, it'll make sure it's installed, and try to load the correct version again. If it fails the second time, it'll print an (error) log statement saying why, and just attempt to load any version it can (and prints out which one it loaded if so).

This also fixes the build issues with Ruby 2.1
### Check List

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -4,18 +4,21 @@ case RUBY_VERSION
 when /^2\.0/
   gem 'chef', '~> 12.0', '< 12.9'
   gem 'berkshelf', '~> 4.3'
+  gem 'foodcritic', '~> 6.3'
 when /^2\.1/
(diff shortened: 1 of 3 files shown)


jsmartt:
Any questions or comments on this one @hdiomede or @tmiotto ? It's pretty basic, but will fix the build issues with the other PRs.

tmiotto (approved):

PR #48, opened 2016-09-29
Server Hardware Type

### LINKED ISSUE #47: Implementation of Server hardware type
### Scenario/Intent

Implement server hardware type actions

hdiomede:
### Description

Server hardware implementation
### Issues Resolved
#47
### Check List


resources/server_hardware_type.rb
```diff
@@ -0,0 +1,28 @@
...
+default_action :edit
+
+action_class do
+  include OneviewCookbook::Helper
+  include OneviewCookbook::ResourceBase
+end
+
+action :edit do
+  item = load_resource
+  add_or_edit(item) if item.exists?
```
tmiotto: If item does not exists it will just pass through without warning the user

tmiotto (changes requested):

resources/server_hardware_type.rb
```diff
@@ -20,7 +20,11 @@
 
 action :edit do
   item = load_resource
-  add_or_edit(item) if item.exists?
+  if item.exists?
+    add_or_edit(item)
+  else
+    Chef::Log.warn "#{resource_name} '#{name}' does not exists"
```
tmiotto: Maybe we should raise an exception. If we are trying to edit an item idempotently, the result should be the item in OneView with the same attributes passed in the recipe, and if we cannot put the item there if it does not exists then we could:
1. Pass through: But this way it loses the idempotent status
2. Raise exception: fails since it is an `unexpected` state

tmiotto (approved):

PR #60, opened 2016-10-03
Implemented action none for interconnect

### LINKED ISSUE #59: Add implementation of action :none to oneview_interconnect
### Scenario/Intent

Default action for interconnects is :none, but it is not implemented.
### Environment Details
- N/A
### Steps to Reproduce

Run any oneview_interconnect resource with default action.
### Expected Result

Nothing should happen.
### Actual Result

it fails since the :none action was not implemented.

tmiotto:
### Description

Implements missing action `:none` for oneview_interconnect
### Issues Resolved

Fixes #59 
### Check List

### DIFF
--- a/examples/interconnect.rb
+++ b/examples/interconnect.rb
@@ -15,6 +15,11 @@ my_client = {
   password: ''
 }
 
+# Do nothing
+oneview_interconnect 'Encl1, interconnect 1' do
+  client my_client
+end
+
 oneview_interconnect 'Encl1, interconnect 1' do
   client my_client
   action :reset_port_protection
--- a/examples/interconnect.rb
+++ b/examples/interconnect.rb
@@ -15,7 +15,7 @@ my_client = {
   password: ''
 }
 
-# Do nothing
+# It will not do anything if no action is selected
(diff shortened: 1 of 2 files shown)


jsmartt:
@tmiotto , just curious why we need the :none action defined? When would someone want to use it like this? (And because it's not clear to me, it would probably be good to add as a comment in the examples)

tmiotto:
Probably no one will want to use the none action, but what would be the default?

We're setting all defaults to create/create_if_missing and variants, but this resource has none of them and there isn't an obvious and eligible action to be the default one, so... It would be best to put a none action, than selecting randomly one of the ones available.

jsmartt:
Ok. Yeah, it seems any action (including :none) would be a little strange. This at least forces the user to pick an action. I don't think we need the example in there to do nothing, but it doesn't hurt anything either I suppose.

jsmartt (approved):

tmiotto:
Since we are using the examples as a kind of documentation, it would be good to warn the user about this behavior there.

I made the message more clear to explain what is happening.

PR #77, opened 2016-10-11
Uplink set

### LINKED ISSUE #76: Uplink Set
### Scenario/Intent

Implementation of Uplink Set actions

anacampesan:
### Description

Implementation of uplink set actions
- added  comments to unmanaged device sample file and increased uplink tests coverage
### Issues Resolved

Fixes #76 
### Check List

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,7 @@
 # Cookbook for HPE OneView
 
 [![Cookbook Version](https://img.shields.io/cookbook/v/oneview.svg)](https://supermarket.chef.io/cookbooks/oneview)
+[![Build Status](https://travis-ci.org/HewlettPackard/oneview-chef.svg?branch=master)](https://travis-ci.org/HewlettPackard/oneview-chef)
 
 Chef cookbook that provides resources for managing OneView.
 
--- a/README.md
+++ b/README.md
@@ -560,11 +560,12 @@ Uplink set resource for HPE OneView
 oneview_uplink_set 'UplinkSet1' do
   client <my_client>
   data <data>
-  native_network # The name of the network that is designated as the native network
-  fc_networks # Array of assigned FC networks
-  fcoe_networks # Array of assigned FCoE networks
-  networks # Array of assigned Ethernet networks
-  logical_interconnect # Name of the assigned Logical Interconnect
+  fc_networks # Array of assigned FC networks - Optional
+  fcoe_networks # Array of assigned FCoE networks - Optional
+  networks # Array of assigned Ethernet networks - Optional
+  logical_interconnect # Name of the assigned Logical Interconnect - Optional
+  native_network # The name of the network that is designated as the native network - Optional
+                 # The native network has to be added to one of the network arrays before being declared
   action [:create, :create_if_missing, :delete]
 end
 ```
--- a/README.md
+++ b/README.md
@@ -557,7 +557,7 @@ end
 Uplink set resource for HPE OneView
 
 ```ruby
-oneview_unmanaged_device 'UnmanagedDevice1' do
+oneview_uplink_set 'UplinkSet1' do
   client <my_client>
   data <data>
   native_network # The name of the network that is designated as the native network
--- a/README.md
+++ b/README.md
@@ -535,6 +535,23 @@ oneview_unmanaged_device 'UnmanagedDevice1' do
 end
 ```
 
+### oneview_uplink_set
+
+Uplink set resource for HPE OneView
+
+```ruby
+oneview_unmanaged_device 'UnmanagedDevice1' do
+  client <my_client>
+  data <data>
+  native_network # The name of the network that is designated as the native network
+  fc_networks # Array of assigned FC networks
+  fcoe_networks # Array of assigned FCoE networks
+  networks # Array of assigned Ethernet networks
+  logical_interconnect # Name of the assigned Logical Interconnect
+  action [:create, :create_if_missing, :delete]
+end
+```
+
 ## Examples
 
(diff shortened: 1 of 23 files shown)


README.md
```diff
@@ -552,6 +552,23 @@ oneview_unmanaged_device 'UnmanagedDevice1' do
 end
 ```
 
+### oneview_uplink_set
+
+Uplink set resource for HPE OneView
+
+```ruby
+oneview_unmanaged_device 'UnmanagedDevice1' do
```
jsmartt: oneview_uplink_set ?

resources/uplink_set.rb
```diff
@@ -0,0 +1,123 @@
...
+    end
+  end
+
+  # Looks for enclosures within Location Entries
+  # An IF statement by itself would try to iterate and raise an exception if locationEntries does not exist
+  def load_enclosure(item)
+    if defined? item.data['portConfigInfos'][0]['location']['locationEntries']
+      item.data['portConfigInfos'][0]['location']['locationEntries'].each do |entry|
+        next unless entry && entry['type'] == 'Enclosure'
+        enclosure = OneviewSDK::Enclosure.find_by(item.client, name: entry['value']).first
```
jsmartt, line 99: Typically this value is a URI. If they specify the URI instead of a name here, this will fail, won't it? Since the retrieve! method tries both, perhaps we can use that instead?

resources/uplink_set.rb
```diff
@@ -0,0 +1,123 @@
...
+action_class do
+  include OneviewCookbook::Helper
+  include OneviewCookbook::ResourceBase
+
+  def load_native_network(item)
+    if native_network
+      if native_network.nil? || native_network == 'nil'
+        item['nativeNetworkUri'] = nil
+      else
+        native_net = OneviewSDK::EthernetNetwork.find_by(item.client, name: native_network).first
```
jsmartt: Is this native network always an EthernetNetwork? The API docs weren't clear on that.

README.md
```diff
@@ -552,6 +552,23 @@ oneview_unmanaged_device 'UnmanagedDevice1' do
...
+  client <my_client>
+  data <data>
+  native_network # The name of the network that is designated as the native network
+  fc_networks # Array of assigned FC networks
+  fcoe_networks # Array of assigned FCoE networks
+  networks # Array of assigned Ethernet networks
```
jsmartt: Please be specific as to what these arrays should include. For example, these are arrays of network **names**, which are strings.

Also, if these properties are optional, please say so.

examples/uplink_set.rb
```diff
@@ -0,0 +1,56 @@
...
+    connectionMode: 'Auto',
+    lacpTimer: 'Short',
+    networkType: 'Ethernet',
+    ethernetNetworkType: 'Tagged',
+    description: nil,
+  )
+  logical_interconnect 'LogicalInterconnect1'
+  networks ['Ethernet1', 'Ethernet2']
+  fc_networks['FCNetwork1']
+  fcoe_network['FCoENetwork1']
```
jsmartt: Seems like there should be a space here (and above). `fcoe_network ['FCoENetwork1']` and `fc_networks ['FCNetwork1']`, right?

examples/uplink_set.rb
```diff
@@ -0,0 +1,56 @@
...
+# CONDITIONS OF ANY KIND, either express or implied. See the License for the
+# specific language governing permissions and limitations under the License.
+
+my_client = {
+  url: '',
+  user: '',
+  password: ''
+}
+
+# Example: creates or updates an uplink set
```
jsmartt, line 25: Maybe everything is self-explanatory (I don't know much about uplink sets), but the more comments here the better. Volumes are pretty simple, but check out the comments on those examples: https://github.com/HewlettPackard/oneview-chef/blob/master/examples/volume.rb#L18

jsmartt (changes requested):

anacampesan:
Thanks for the insights, all the issues have been addressed/commented in the code. Could you please take another look, @jsmartt? Ty!

resources/uplink_set.rb
```diff
@@ -0,0 +1,134 @@
...
+# CONDITIONS OF ANY KIND, either express or implied. See the License for the
+# specific language governing permissions and limitations under the License.
+
+OneviewCookbook::ResourceBaseProperties.load(self)
+
+property :native_network, String       # Native network name - can be null
+property :networks, Array              # Array of strings containing Ethernet Network names - required
+property :fc_networks, Array           # Array of strings containing FC network names - required
+property :fcoe_networks, Array         # Array of strings containing FCoE Network names - required
+property :logical_interconnect, String # Associated logical interconnect name - required
```
jsmartt: Everything else looks good, but for these, they have "required" in the comments here, but there's `if` statements below that seem to be perfectly happy to skip these attributes if they're not set. If they are actually required, I think we need to have the `required: true` [directive](https://docs.chef.io/custom_resources.html#define-properties) here, not just a comment.

However, I think it's probably best to just keep them optional so users can set these things in the data hash if they want to. If they're optional, I think it would be good to say so here in these comments, but also in the README. Like [this](https://github.com/HewlettPackard/oneview-chef#oneview_server_hardware)

anacampesan: I figured out none of them are really necessary, as the default values are being set by the SDK. The only one that is not is nativeNetworkUri, which I manually set it as 'nil' in case it has not been added by the used. Please review it again! @jsmartt

jsmartt:
Shoot. Well, trying to push a change added some recent commits from master. The only real changes I made were on the "Updates" commit. But I'm good with this now :smile:

jsmartt (approved):

PR #89, opened 2016-11-02
Added Ruby version to Gemfile

### LINKED ISSUE #85: Fix minimum Ruby lib dependency
### Scenario/Intent
Some dependencies require lower versions of Ruby and we need to manage them so Travis won't fail.

jsmartt:
### Description
Adds the current Ruby version to the Gemfile so that dependencies are solved, taking into consideration the current Ruby version. Don't know why this was so hard to find, but I finally ran across [bundler PR 4651](https://github.com/bundler/bundler/pull/4651) that described this workaround.

### Issues Resolved
Fixes #85

### Check List
N/A. Just an improvement for dev

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,26 +1,10 @@
 source 'https://rubygems.org'
 
-case RUBY_VERSION
-when /^2\.0/
-  gem 'chef', '~> 12.0', '< 12.9'
-  gem 'berkshelf', '~> 4.3'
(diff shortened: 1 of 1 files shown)


jsmartt:
Any questions/concerns with this @tmiotto or @hdiomede ?

tmiotto (approved):

PR #96, opened 2017-01-05
Changes for Synergy support

tmiotto:
### Description
Adds architectural support to Synergy and API300 resources

:warning: There is 1 expected failure: (See oneview-sdk-ruby's [#124 Missing argument in API300 C7000 Managed SAN method](https://github.com/HewlettPackard/oneview-sdk-ruby/issues/124) for more information).

Regarding *dynamic version load*:
It won't load SDK version and variants at converge time by now. It would be better to open an issue/another PR until we have any good idea on how to kick it in, since it won't be a big deal to refactor and the implementation may be delayed if we stop the development to think about it right now.

### Main changes ###
- Added berkshelf exception to ignore
- Excluded libraries folder from foodcritic
- Addded load_attributes method
- Converted load_sdk to static method
- Added structure in libraries to receive providers
- Implemented EthernetNetwork resource as example
- Edited specs and resources to accept the new standard

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,6 +10,7 @@ pkg/
 .vagrant
 /cookbooks
 Berksfile.lock
+.berkshelf/
 
 # Bundler
 Gemfile.lock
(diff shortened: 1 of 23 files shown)


tmiotto:
No need for this anymore since the changes made here were merged into #99

PR #99, opened 2017-01-10
Synergy support option 2

### LINKED ISSUE #97: Support multiple API versions
### Scenario/Intent
Support more API versions and the new/updated resources associated with them

### Environment Details
 - **OneView Version:** 200, 300, and beyond

### Description
In order to support newer API versions and yet maintain compatibility with previous ones, we'll need to do a pretty big refactor. The work is already underway, but this issue can capture some of the ideas and thoughts behind it, and serve as a discussion forum. From what we've discussed already, this refactor will likely include:
 - Splitting most of the logic for resource actions into separate "provider" modules/classes.
 - Supporting only 1 API version and HW variant per chef-client run. Because of the differences in the resources for separate API versions and how the providers would have to be split out, we need to load the providers at compile time, not runtime.

### Outstanding Questions
1. How do we handle resources that have new/updated properties for newer API versions? For example, for `oneview_server_profile`, how do we add an os_deployment_image property to the API300 provider but make it clear that it isn't used in the API200 provider? Seems like there's a lot of room for confusion here.
2. Similarly, how should completely new resources be handled that don't exist in API200 (e.g., all the I3S stuff)?
3. Is there a good way to support multiple versions and variants in a chef-client run? I haven't found a good solution for it yet, but it may be possible.

jsmartt:
### Description
This PR offers another option for addressing support for Synergy and future OV releases. (At this point I just have the ethernet_network providers done.) The logic for resources is split up into separate providers for each API version and variant. Here are the basic features:
- Run-time loading of the correct SDK module, variant, and class.
- Can use multiple SDK variants & API versions in a single Chef run
- Can override the X-API-Version header in API requests by setting a property on a resource
- Basically all the logic is split out into Provider **classes**, not modules
- Provider classes expose many of the helper methods have been copied into the provider class so they can be easily used where they will be needed

Compared to #96, this option will require more work to refactor, but I think it's worth it.

### Issues Resolved
Fixes #97

### Check List

@tmiotto or others, let me know what you think!


libraries/resources/api200/resource_provider.rb
```diff
@@ -0,0 +1,21 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software distributed
+# under the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR
+# CONDITIONS OF ANY KIND, either express or implied. See the License for the
+# specific language governing permissions and limitations under the License.
+
+module OneviewCookbook
+  module API200
+    # Base class for API200 resource providers
+    class EthernetNetwork < OneviewCookbook::ResourceProvider
```
tmiotto: EthernetNetwork? I don't get it

jsmartt: Oops. Should be fixed now.

libraries/resources/api200/resource_provider.rb
```diff
@@ -0,0 +1,21 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software distributed
+# under the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR
+# CONDITIONS OF ANY KIND, either express or implied. See the License for the
+# specific language governing permissions and limitations under the License.
+
+module OneviewCookbook
+  module API200
+    # Base class for API200 resource providers
+    class ResourceProvider < OneviewCookbook::ResourceProvider
```
tmiotto: Ok, but why do we have this?

It is to add any shared behavior specific to API200? If so, shouldn't we have to do something similar to API300 or C7000 and Synergy?

jsmartt: Yes. I didn't include the one for API300 and the variants YET, but yes, the concept is the same. We'd add the default class in there. Exactly like the Resource class for the SDK

libraries/.foodcritic
```diff
@@ -0,0 +1,4 @@
+~FC011
```
tmiotto: I guess foodcritic is failing because we're running it from the wrong directory.

Actually, it should receive as parameter the `cookbooks/` folder, so it gets our `oneview` cookbook and then analyzes it. But in this case it is being executed inside `oneview/`, so this way it is interpreting the `libraries` as a separate cookbook, and not a subfolder

tmiotto: Well, it's not exactly the *wrong directory*, but running on `cookbooks/` would solve the problem (but this is a workaround), or maybe we should just add an `exclude_path` if we don't want libraries to be analyzed by it.

jsmartt: I can't get it to **not** see the libraries directory as a separate cookbook, which is why I put that `.foodcritic` file in there. How are you able to run it so it gets ignored?

jsmartt: The travis test doesn't have a `cookbooks` directory. Just this repo.

jsmartt: I think the real issue is that foodcritic thinks the libraries directory is a cookbook because of the `resources` directory within it. We can rename the directory to avoid these issues. See [this issue](https://github.com/acrmp/foodcritic/issues/148)

jsmartt: Should be fixed now

tmiotto:
[These changes](https://github.com/HewlettPackard/oneview-chef/pull/96/commits/56dd9f2867b917c6c5c7e5c2ddbd4084c9bb616d) from the PR #96 should be added to stop the spec from failing.

jsmartt:
Added some more tests and functionality. I think the functionality for EthernetNetworkProvider is now done. Just need a few more test, but I'll get to those tomorrow.

libraries/resource_providers/api300.rb
```diff
@@ -20,7 +20,7 @@ module API300
     # @return [Class] Resource class or nil if not found
     def self.resource_named(type, variant)
       raise "API300 variant #{variant} is not supported!" unless SUPPORTED_VARIANTS.include?(variant.to_s)
-      new_type = type.to_s.downcase.gsub(/[ -_]/, '')
+      new_type = type.to_s.downcase.gsub(/[ -_]/, '') + 'provider'
```
tmiotto, line 23: Should we still be calling it `resource_named`?

spec/unit/resources/ethernet_network/create_spec.rb
```diff
@@ -27,3 +27,14 @@
     expect(real_chef_run).to create_oneview_ethernet_network('EthernetNetwork1')
   end
 end
+
```
tmiotto, line 30: Maybe we would not want to implement tests of things that are already being tested on API200

libraries/resource_providers/api300.rb
jsmartt: What do you suggest instead? `resource_provider_named` ?

tmiotto: It is an option, or maybe just `provider_named`.

spec/unit/resources/ethernet_network/create_spec.rb
jsmartt: Agreed. This is just a sample of what you would do if you wanted to test API300 resources. I think this particular one is probably OK to leave in since it's the only one so far, but we wouldn't have to add these sort of tests for other resources that are the same as API200.

tmiotto:
@jsmartt Just one more thought...

As far as we go through the implementation, we'll always be using the last API version as base for the next one, i.e. `API300::X < API200::X`, `API500::X < API300::X`.

Said that we won't be using any ResourceProvider other than the `OneviewCookbook::ResourceProvider` since `API500::X < API300::X < API200::X < OneviewCookbook::ResourceProvider`, and the ones created aren't adding anything new.

It won't be better to just have the `OneviewCookbook::ResourceProvider`?

jsmartt:
Yeah, I guess I added the API200 sub-class prematurely. We don't need it now, so there's no need to include it. I'll remove it

jsmartt:
OK @tmiotto , I'm ready on this when you are 😄

tmiotto (approved):

PR #104, opened 2017-01-12
Fabrics resource API300 support

### LINKED ISSUE #103: API300 support to Fabrics
### Scenario/Intent
Add support to the new Fabrics functionality `:set_reserved_vlan_range`

- [ ]  Create the resource Fabrics
- [ ]  Create action `:set_reserved_vlan_range` (**API300::Synergy only**)
- [ ]  Add tests for Fabrics

tmiotto:
### Description
Adds new resource, tests and support to Fabrics.

Compatible only with API300::Synergy.

### Issues Resolved
#103 

### Check List

### DIFF
--- a/README.md
+++ b/README.md
@@ -93,6 +93,23 @@ oneview_ethernet_network 'Eth1' do
 end
 ```
 
+### oneview_fabric
+
+Fabric resource for HPE OneView.
+
+Support only in API300 onwards with variant Synergy.
+
+Action defaults to `:none`, also performs updates on the reserved vlan range.
+
+```Ruby
+oneview_fabric 'Fabric1' do
+  client <my_client>
+  data <resource_data>
+  reserved_vlan_range <vlan_options> # Hash: Usually the 'start' and 'length' of the range
+  action [:none, :set_reserved_vlan_range]
+end
+```
+
 ### oneview_fc_network
 
 FC network resource for HPE OneView.
@@ -106,7 +123,7 @@ end
 ```
 
 
-### oneview_fc_network
+### oneview_fcoe_network
 
 FCoE network resource for HPE OneView.
 
--- a/README.md
+++ b/README.md
@@ -99,14 +99,14 @@ Fabric resource for HPE OneView.
 
 Support only in API300 onwards with variant Synergy.
 
-Action defaults to `:none`, also performs updates on the reserved vlan range.
+Performs updates on the reserved vlan range.
 
 ```Ruby
 oneview_fabric 'Fabric1' do
   client <my_client>
   data <resource_data>
   reserved_vlan_range <vlan_options> # Hash: Usually the 'start' and 'length' of the range
-  action [:none, :set_reserved_vlan_range]
+  action [:set_reserved_vlan_range]
 end
 ```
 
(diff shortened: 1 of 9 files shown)


examples/fabric.rb
```diff
@@ -0,0 +1,31 @@
...
+
+# Updates the DefaultFabric fabric reserved vlan range if the values do not match
+oneview_fabric 'DefaultFabric' do
+  client my_client
+  reserved_vlan_range(
+    'start' => 3000,
+    'length' => 120
+  )
+  action :set_reserved_vlan_range
+end
```
jsmartt, line 28: Because the default API version is 200 and the default variant is C7000, it would be helpful to explicitly set those properties here.

README.md
```diff
@@ -93,6 +93,23 @@ oneview_ethernet_network 'Eth1' do
 end
 ```
 
+### oneview_fabric
+
+Fabric resource for HPE OneView.
+
+Support only in API300 onwards with variant Synergy.
+
+Action defaults to `:none`, also performs updates on the reserved vlan range.
```
jsmartt: Just curious, why the default action of `:none`?

jsmartt (changes requested):

README.md
tmiotto: Well, since the default is always `:create` or `:none` and the resource has no `:create` action, I put `:none`

examples/fabric.rb
tmiotto: @jsmartt 

I haven't seen the code loading the api_version and variant from the client. Is it doing it?

jsmartt: @tmiotto , no. It doesn't use the client, but it does use the optional properties `api_version` and `api_variant`; see [here](https://github.com/HewlettPackard/oneview-chef/blob/master/libraries/oneview_helper.rb#L33). Those default to the attributes defined by the cookbook; see [here](https://github.com/HewlettPackard/oneview-chef/blob/master/libraries/oneview_resource_base.rb#L21) and [here](https://github.com/HewlettPackard/oneview-chef/blob/master/attributes/default.rb#L27).

So based on the example code I see, it would use the cookbook defaults: `api_version=200` & `api_variant='C7000'`. Even if we allow it to use the client's version, it wouldn't be able to figure out the variant. And the version isn't specified in this example anyways.

README.md
jsmartt: Do you envision a user ever using the `:none` action? I know it can be useful for cases when you want to notify it from another resource, but I don't know enough about fabrics to know if that's a valid use case for this resource.

tmiotto: Yeah, probably not...

README.md
```diff
@@ -93,6 +93,23 @@ oneview_ethernet_network 'Eth1' do
...
+
+```Ruby
+oneview_fabric 'Fabric1' do
+  client <my_client>
+  data <resource_data>
+  reserved_vlan_range <vlan_options> # Hash: Usually the 'start' and 'length' of the range
+  action [:none, :set_reserved_vlan_range]
```
jsmartt: After you remove the 2 references to the `:none` action in the README (this one and the one above), I'm good with this :)

jsmartt (changes requested):

fgbulsoni (approved):
Looks good. The other changes requested by @jsmartt seem to be fixed too. :+1:

jsmartt (approved):

PR #183, opened 2017-02-07
README updates

jsmartt:
### Description
README updates that were forgotten for the 1.2.0 release

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,7 +9,7 @@ Chef cookbook that provides resources for managing HPE OneView.
 ## Requirements
 
  - Chef 12.0 or higher
- - HPE OneView 2.0 or 3.0 (may work with other versions too, but no guarantees)
+ - HPE OneView 2.0 or 3.0 (API versions 200 or 300). May work with other versions too, but no guarantees
 
 ## Usage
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -4,33 +4,34 @@
...
 [![Chef Build Status](https://jenkins-01.eastus.cloudapp.azure.com/job/oneview-cookbook/badge/icon)](https://jenkins-01.eastus.cloudapp.azure.com/job/oneview-cookbook/)
 
-Chef cookbook that provides resources for managing OneView.
+Chef cookbook that provides resources for managing HPE OneView.
 
 ## Requirements
 
  - Chef 12.0 or higher
- - HPE OneView 2.0 (may work with other versions too, but no guarantees)
+ - HPE OneView 2.0 or 3.0 (may work with other versions too, but no guarantees)
```
tmiotto: It would be better to specify also that actually the support focuses on the REST API200 and API300

jsmartt: How about
> HPE OneView appliance that supports API version 200 or 300 (may work with other versions too, but no guarantees)

This may make it a little more difficult for people that don't know what API version they have, but do know the release number (i.e., 2.0)

tmiotto: My idea is to keep the HPE OneView release versions and add also the REST API versions

PR #189, opened 2017-02-16
Image Streamer Refactor

tmiotto:
### Description
 - Refactors the 1.3.0 code to support the addition of secondary APIs
 - Introduces the new ImageStreamer::API300 API
 - Adds the resource image_streamer_plan_script (OneviewSDK::ImageStreamer::API300::PlanScript)
 - Specs supporting both ImageStreamer and OneView resources

⚠️  ~SimpleCov coverage not working for some underworldly reason.~ (If any ChefSpec is executed first than the pure RSpec ones, SimpleCov simply halts and outputs the coverage)
⚠️  ~Berkshelf cannot find the compat_resource cookbook in Travis build~ (For some reason it won't work if we call rake using bundle)

### Check List


.travis.yml
```diff
@@ -2,9 +2,10 @@ language: ruby
 rvm:
   - 2.1
   - 2.2.5
+  - 2.2.6
```
jsmartt: I think we can just get rid of 2.2.5 if we add this version.

.travis.yml
```diff
@@ -2,9 +2,10 @@ language: ruby
...
   - 2.3.1
 
 before_install: gem update bundler
 
 script:
-  - rake test
+  - bundle exec rake test
```
jsmartt: rake should already load bundler, but did you find this was necessary?

attributes/default.rb
```diff
@@ -24,7 +24,7 @@
 # A resource provider must be defined for this version. For example, when set to 200, it will look
 # for the resource in OneviewCookbook::API200. When 300, it will look in OneviewCookbook::API300
 # See the libraries/resources directory for more info on supported API versions
-default['oneview']['api_version'] = 200
+default['oneview']['api_version'] = 300
```
jsmartt: This will require a major version bump (breaking change). That's fine, just wanted to make note of it.

.rubocop.yml
```diff
@@ -14,7 +14,7 @@ Metrics/LineLength:
 
 Metrics/MethodLength:
   CountComments: false
-  Max: 26
+  Max: 50
```
jsmartt: 50 lines really is pretty long for a method. Maybe we can split the offending method apart a bit?

libraries/resource_provider.rb
```diff
@@ -19,35 +19,59 @@ class ResourceProvider
...
+          end
+      new_data = JSON.parse(context.data.to_json) rescue context.data
+      @item = context.property_is_set?(:api_header_version) ? klass.new(c, new_data, context.api_header_version) : klass.new(c, new_data)
+      @item['name'] ||= context.name
+    end
+
+    # Helper method that analyzes the namespace and defines the class variables and the resource class itself
+    # @return [OneviewSDK::Resource] SDK resource class
+    # @raise [NameError] If for some reason the method could not resolve the namespace to a OneviewSDK::Resource
+    def namespace_handler
```
jsmartt: And just as a suggestion, maybe we can use a more descriptive name for this method. Something like `load_namespace`, `load_class_from_namespace` or `parse_namespace`? Handler makes it sound like an exception handler or something, but this method both sets data and returns important info (class reference).

jsmartt (commented):
There's a lot in this PR, so I'll take me a while to look through it all, but here's a few initial comments without digging in too much.

attributes/default.rb
tmiotto: No need for this change, I pushed it accidentally

.rubocop.yml
tmiotto: Actually, the rubocop is skipping the comments, but not the log messages, tha is why I had to make it bigger.

Do you know if it is possible to add this kind of exception to rubocop?

The `namespace_handler`or `parse_namespace` has a lot of log lines that are being accounted as normal code lines.

I can bump it down back to the previous value, but it would cause some methods to be splitted unnecessarily.

.travis.yml
tmiotto: Since Travis starts is in a clean state it won't be needed

.travis.yml
tmiotto: Yep, both are there just for validation purposes before choosing one option over another

.rubocop.yml
jsmartt: If it doesn't make sense to split them, this is fine. I'm not aware of any way to filter out logging lines.

fgbulsoni: If this issue is specific to this method due to all the logging it is doing and that is expected (which looks like it), plus it does not make sense to split it, we might just want to disable the cop for this one method instead of increasing the threshold for all methods in the project.
By this time it was already refactored into two methods, but just pointing out that this might sometimes be the more readable approach.
To do it we'd just need to add `# rubocop:disable Metrics/MethodLength` to the method `def` line.

fgbulsoni (approved):
Big PR :scream_cat: But overall looks real good, specially with the changes commented by @jsmartt. :shipit: !

tmiotto:
@jsmartt Any other suggestions here, or is it ready to merge?

(Notice it is going to be merged on `master_i3s` and not `master`)

jsmartt:
Sorry @tmiotto , we had a holiday yesterday in the US, so I'm just now getting around to giving it a look. Mainly trying to figure out how the `sdk_secondary_api` parts are used. Should be done in an hour or so

libraries/resource_provider.rb
```diff
@@ -19,35 +19,67 @@ class ResourceProvider
       :item,          # The OneviewSDK resource instance
       :sdk_api_version,  # Keep track of the API version used for the SDK resource class
       :sdk_variant,      # Keep track of the variant used for the SDK resource class
-      :sdk_resource_type # Keep track of the SDK resource class name
+      :sdk_resource_type, # Keep track of the SDK resource class name
+      :sdk_secondary_api  # Keep track of the internal secondary API used by the SDK
```
jsmartt: This should probably be an `attr_reader`. And like the other comments below, how about naming it `base_module` and setting a default (`OneviewSDK`)?

libraries/oneview_helper.rb
```diff
@@ -55,10 +60,12 @@ def self.load_sdk(context)
     rescue LoadError => e
       Chef::Log.debug("Could not load gem oneview-sdk #{node['oneview']['ruby_sdk_version']}. Message: #{e.message}")
       Chef::Log.info("Could not load gem oneview-sdk #{node['oneview']['ruby_sdk_version']}. Making sure it's installed...")
+      # :nocov:
```
jsmartt: Why add the nocov comment here? I think it's better to have the feedback that it's not covered than removing it from the report.

libraries/oneview_helper.rb
```diff
@@ -18,8 +18,8 @@ module Helper
     # @param context Context from the resource action block (self)
     # @param [String, Symbol] type Name of the resource type. e.g., :EthernetNetwork
     # @param [String, Symbol] action Action method to call on the resource. e.g., :create_or_update
-    def self.do_resource_action(context, type, action)
-      klass = OneviewCookbook::Helper.get_resource_class(context, type)
+    def self.do_resource_action(context, type, action, secondary_api = nil)
```
jsmartt: All these references to 'secondary_api' are pretty confusing to me, partly because when it's set, it's actually the only api module, not a secondary one. How about we name it something 'base_module'? And I think it makes the most sense to set the default to `OneviewCookbook` instead of nil. That way it's clear what the default is. Also, add the `@param` docs describing it and accepted value(s).

libraries/oneview_helper.rb
```diff
@@ -28,21 +28,26 @@ def self.do_resource_action(context, type, action)
     # @param context Context from the resource action block (self)
     # @param [String, Symbol] type Name of the resource type. e.g., :EthernetNetwork
     # @return [Class] Resource class
-    def self.get_resource_class(context, resource_type)
+    def self.get_resource_class(context, resource_type, secondary_api = nil)
```
jsmartt: Same here with the param name, docs, and default

libraries/oneview_helper.rb
```diff
@@ -28,21 +28,26 @@ def self.do_resource_action(context, type, action)
...
     # @return [Module] Resource module
-    def self.get_api_module(api_version)
-      OneviewCookbook.const_get("API#{api_version}")
+    def self.get_api_module(api_version, secondary_api = nil)
+      base_sdk = secondary_api ? OneviewCookbook.const_get(secondary_api) : OneviewCookbook
+      base_sdk.const_get("API#{api_version}")
     rescue NameError
-      raise NameError, "The api_version #{api_version} is not supported. Please use a supported version."
+      msg = "The api_version #{api_version} is not supported"
+      msg += secondary_api ? " in #{secondary_api}" : '.'
```
jsmartt: Same here with the param name, docs, and default. This will simplify the method too.

libraries/resource_provider.rb
```diff
@@ -19,35 +19,67 @@ class ResourceProvider
...
+      end
+      base_sdk = @sdk_secondary_api ? OneviewSDK.const_get(@sdk_secondary_api) : OneviewSDK
+      klass ||= base_sdk.resource_named(@sdk_resource_type, @sdk_api_version, @sdk_variant)
+      klass
+    end
+
+    # Parses all the namespace whrn it is fully specified with API version, variant or secondary APIs
+    # @param [String] klass_name Name of the provider class
+    # @param [Array<String>] name_arr Remaining namespace list that may contain variant, api version and secondary api name
+    def parse_namespace_with_secondary_apis(klass_name, name_arr)
```
jsmartt: I think all this should be included in the method above. If we need to disable the method-size rubocop rule with comments around it that's fine, but splitting it is confusing.

jsmartt (changes requested):

libraries/resource_provider.rb
tmiotto: Great, I had was the same thought on naming it `base_module` but I guessed it could be confusing naming it this way... but it's prettier... I'll go with it.

libraries/resource_provider.rb
tmiotto: Agreed

libraries/oneview_helper.rb
tmiotto: the nocov is here because I checked the specs and, it's clearly passing through here, but simplecov cannot detect it executed the block inside the chef_gem, although being confirmed it did.

tl;dr Simplecov cannot detect this thing inside being executed, but it's being covered.

jsmartt: It passes **through** here, but it doesn't step **inside** the chef_gem block by default because in chefspec, it basically runs in --why-run mode, meaning only the resources that we say it should step into get executed. This is why we do [this](https://github.com/HewlettPackard/oneview-chef/blob/master/spec/spec_helper.rb#L58). I tested it out by putting a `puts` statement inside this block and running the tests, and it never printed anything. And I think it's OK that it doesn't step into it. I think there's even a test somewhere that mocks it out.

tmiotto: Oh the `--why-run`... Alright so, I'll remove

tmiotto:
@jsmartt I improved it, could you take a look?

spec/spec_helper.rb
```diff
@@ -30,8 +29,12 @@
     allow_any_instance_of(OneviewSDK::Client).to receive(:login).and_return('secretToken')
     allow(OneviewSDK::SSLHelper).to receive(:load_trusted_certs).and_return(nil)
 
+    # Mock Image stremaer appliance version
+    allow_any_instance_of(OneviewSDK::ImageStreamer::Client).to receive(:appliance_i3s_api_version).and_return(300)
+
     # Clear environment variables
-    %w(ONEVIEWSDK_URL ONEVIEWSDK_USER ONEVIEWSDK_PASSWORD ONEVIEWSDK_TOKEN ONEVIEWSDK_SSL_ENABLED).each do |name|
+    %w(ONEVIEWSDK_URL ONEVIEWSDK_USER ONEVIEWSDK_PASSWORD ONEVIEWSDK_TOKEN ONEVIEWSDK_SSL_ENABLED
```
jsmartt: Should we just use `OneviewSDK::ENV_VARS` here instead? That way it keeps up to date automatically. (e.g., we just got rid of I3S_TOKEN)

libraries/oneview_helper.rb
```diff
@@ -18,31 +18,39 @@ module Helper
...
+    def self.get_resource_class(context, resource_type, base_module = OneviewCookbook)
       load_sdk(context)
-      api_version = context.property_is_set?(:api_version) ? context.api_version : context.node['oneview']['api_version']
-      api_module = get_api_module(api_version)
+      # Loads the api version if the property was specified directly
+      context_api_version = context.api_version if context.property_is_set?(:api_version)
+      # Loads the api version if it was specified in the client
+      client_api_version = convert_keys(context.client, :to_sym)[:api_version] if context.property_is_set?(:client)
+      # Loads the api version giving preference: property > client > node default
+      api_version = context_api_version || client_api_version || context.node['oneview']['api_version']
```
jsmartt: This is a small but significant change that I missed before. I hope it doesn't affect users negatively. Ideally the API versions for the client, and module will match, but I'm sure that's not always the case. Especially if the user doesn't declare an API version, so it gets set to `OneviewSDK::DEFAULT_API_VERSION` (200). Even if they set `node['oneview']['api_version']` to 300, for example, it would look for API200 resources. Personally I'd keep the behavior the way it was before. Any thoughts on this @tmiotto ?

(If you want to keep it this way, we need to add a note in the attributes file and README explaining the order of precedence)

tmiotto: @jsmartt  The way it was being defined before was completely ignoring the Client API version.

Ideally, if no property is specified saying which API version should be used, the client API version must be used.

I think it was counterintuitive setting the client API version as 300, and then executing the API200, because it was the node default, but the preference was being overriden by the client.

fgbulsoni: > Especially if the user doesn't declare an API version, so it gets set to OneviewSDK::DEFAULT_API_VERSION (200).

So, my 10 cents on this as a new user of the module:

I actually expected it to be using the API I had declared for the client. 
At first, I was trying to just declare the client and call my resources, and I was hitting errors because the resource did not exist for version 200, and that was something I felt was really counterintuitive. 
I then had to actually declare in each resource the version I wanted it to use, instead of it 'in a smart way' getting the version from the client I was passing in.
Bottom line: as a new user, I'd expect it the api version to match my client, and if I actually want to differ from that client version, I'd then explicitly tell the resource which api version to use, not the other way around.

jsmartt: If that's the case, I think we need to remove the `node['oneview']['api_version']` attribute altogether then. Because from what I can tell, the client's api_version will always be set. It'll never get to that final `||=`. Correct?

In the case that there is no client property given, then we can set a default I guess, but I don't think it needs to be an attribute, and I can't think of a time that it'll ever even happen.

jsmartt: Either way, I think this needs to be considered a breaking change, because that attribute will be all but ignored.

tmiotto: Not exactly ignored. It will get the client version before the client is built.

So the client will only have api_version by default if it is a `OneviewSDK::Client`, and if the user is creating directly the `OneviewSDK::Client` he must know the implications.

I agree it would be a breaking change, but I guess it won't be a problem.

jsmartt: Good point. I'm OK with this as long as you put a note about it in the attributes file and main README.

jsmartt:
Also, we need to update the README with all the new I3S stuff. Like the additional ENV var(s) [here](https://github.com/HewlettPackard/oneview-chef#credentials), client types, etc.

fgbulsoni (changes requested):
Since `plan_script` is being added in this PR as well, an `example` for it is missing and should probably be added

tmiotto:
@fgbulsoni @jsmartt Added the last bits. See if I missed something or if you have any further suggestions

fgbulsoni (approved):

README.md
```diff
@@ -84,7 +104,7 @@ end
 
 See the [example](examples/oneview_resource.rb) for more details.
 
-### oneview_ethernet_network
+#### oneview_ethernet_network
```
jsmartt: I like the idea of organizing resources by type, but I don't think we can make the resource name headings any smaller than H3. When it's H4 like this, resources like [oneview_logical_interconnect_group](https://github.com/HewlettPackard/oneview-chef/blob/7c3b482d3667ad925c3172d245df751c9dbeaf3d/README.md#oneview_logical_interconnect_group) get pretty confusing, as some of the properties are the same size as the resource heading. You can always bump up the size of `Resources` and `Resources` to H1 and H2.

Or you can just remove the OneView & ICSP organizational headings. It will be pretty clear which are which.

libraries/oneview_helper.rb
```diff
@@ -35,7 +35,10 @@ def self.get_resource_class(context, resource_type, base_module = OneviewCookboo
       # Loads the api version if the property was specified directly
       context_api_version = context.api_version if context.property_is_set?(:api_version)
       # Loads the api version if it was specified in the client
-      client_api_version = convert_keys(context.client, :to_sym)[:api_version] if context.property_is_set?(:client)
+      client_api_version = if context.property_is_set?(:client)
+                             context.client = convert_keys(context.client, :to_sym) if context.client.is_a?(Hash)
+                             context.client[:api_version]
```
jsmartt: Won't this fail if the client is a `OneviewSDK::Client` object?

libraries/oneview_helper.rb
```diff
@@ -18,31 +18,42 @@ module Helper
...
+    def self.get_resource_class(context, resource_type, base_module = OneviewCookbook)
       load_sdk(context)
-      api_version = context.property_is_set?(:api_version) ? context.api_version : context.node['oneview']['api_version']
-      api_module = get_api_module(api_version)
+      # Loads the api version if the property was specified directly
+      context_api_version = context.api_version if context.property_is_set?(:api_version)
+      # Loads the api version if it was specified in the client
+      client_api_version = if context.property_is_set?(:client)
+                             context.client = convert_keys(context.client, :to_sym) if context.client.is_a?(Hash)
+                             context.client[:api_version]
```
jsmartt: Won't this fail if the client is a `OneviewSDK::Client` object?

CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 2.0.0 - Image Streamer
+Adds support to API300 HPE Synergy Image Streamer resources:
+  - image_streamer_plan_script
```
jsmartt: Add note here saying what the breaking change(s) are

jsmartt (changes requested):
After these two things, I think it's ready

jsmartt:
^^ 3 things, not 2

jsmartt (approved):
OK, it's ready to go now I think 👍

PR #196, opened 2017-03-02
Add support to image streamer deployment plan

### LINKED ISSUE #193: Add support to HPE Image Streamer's Deployment Plans
### Scenario/Intent
Use the HPE Image Streamer's Deployment Plans resource that's been added in the oneview-ruby-sdk

### Environment Details
 - **oneview Cookbook Version:** 1.3.0
 - **OneView Version:** 3.0.0
 - **OneView API Version:** 300
 - **chef-client Version:** 12.18.31

### Steps to Reproduce
Try to use the image streamer resource Deployment Plans through the oneview cookbook

### Expected Result
It should have the resource and be possible to use it

### Actual Result
This is currently not implemented.

fgbulsoni:
### Description
Adds the `image_streamer_deployment_plan` resource to the oneview-cookbook

### Issues Resolved
#193 

### Check List
- [X] New functionality includes testing.
  - [X] All tests pass (`$ rake test`).
- [X] New functionality has been documented in the README if applicable.
  - [X] New functionality has been thoroughly documented in [examples](https://github.com/HewlettPackard/oneview-chef/tree/master/examples/) (please include helpful comments).
- [X] Changes documented in the [CHANGELOG](https://github.com/HewlettPackard/oneview-chef/blob/master/CHANGELOG.md).


examples/image_streamer/deployment_plan.rb
```diff
@@ -0,0 +1,49 @@
...
+
+# Update the Deployment Plan named 'DeploymentPlan1' with a new description
+image_streamer_deployment_plan 'DeploymentPlan1' do
+  client i3s_client
+  api_version 300
+  data(
+    description: 'Chef created and updated Deployment Plan - 1'
+  )
+end
+
```
tmiotto, line 43: Create the `create_if_missing` action if missing

libraries/resource_providers/image_streamer/api300/deployment_plan.rb
```diff
@@ -0,0 +1,44 @@
...
+  module ImageStreamer
+    module API300
+      # DeploymentPlan Provider resource methods
+      class DeploymentPlanProvider < ResourceProvider
+        # Generic method to retrieve and return the URI from different image streamer resources using the name of the resource
+        # @param [String, Symbol] resource_field Name of the field which will be receiving a resource name. e.g., 'golden_image'
+        def load_i3s_resource(resource_field)
+          return unless @context.method(resource_field).call
+          resource_name = @context.method(resource_field).call
+          klass_name = resource_field.to_s.split('_').collect(&:capitalize).join
```
tmiotto: Why are you doing it?

You simply could have used something like `load_i3s_resource(:GoldenImage)` or `load_i3s_resource('GoldenImage')` and would have avoided all this unnecessary inflate.

tmiotto (changes requested):
Also missing the update in the matcher_spec.rb

libraries/resource_providers/image_streamer/api300/deployment_plan.rb
fgbulsoni: Addiction to inflate and reducing stuff passed in to a method I guess. :mask: 
Changing according to requested

libraries/resource_providers/image_streamer/api300/deployment_plan.rb
```diff
@@ -0,0 +1,43 @@
...
+
+module OneviewCookbook
+  module ImageStreamer
+    module API300
+      # DeploymentPlan Provider resource methods
+      class DeploymentPlanProvider < ResourceProvider
+        # Generic method to retrieve and return the URI from different image streamer resources using the name of the resource
+        # @param [String, Symbol] i3s_resource_type Type of image streamer resource to be retrieved. e.g., ':GoldenImage'.
+        # @param [String] i3s_resource_name Name of the resource from context.
+        def load_i3s_resource(i3s_resource_type, i3s_resource_name)
```
fgbulsoni: @tmiotto @jsmartt Question (Maybe for the future?): Would it be useful to port this method from this resource into something of a larger scope, like the `resource_provider` or something else? 
I'm imagining there will be other very similar methods and usecases across the i3s at least.

tmiotto: I have one question...

Is it worth to create this method to replace:

This:
```ruby
data['myTypeUri'] = resource_named(:MyType).find_by(name: @context.my_type) if @context.my_type
```

For this?
```ruby
data['myTypeUri'] = load_i3s_resource(:MyType, @context.my_type) if @context.my_type
```

README.md
```diff
@@ -877,6 +876,20 @@ end
...
 
+### image_streamer_deployment_plan
+
+HPE Synergy Image Streamer resource for Deployment plans.
+
+```ruby
+image_streamer_deployment_plan 'DeploymentPlan1' do
+  client <my_client>   # Hash or OneviewSDK::ImageStreamer::Client
+  data <resource_data> # Hash
+  build_plan <os_build_plan_name> # String containing the name of the OS Buld Plan to be associated to this deployment plan - Optional
```
tmiotto: name of the OS _Bu**i**ld_ Plan

libraries/resource_providers/image_streamer/api300/deployment_plan.rb
fgbulsoni: I'd usually like to have a raise in case a resource I'm passing in does not exist, but if it's all the same, I can drop the method and just swap out the method calls for the aforementioned `data['myTypeUri'] = resource_named(:MyType).find_by(name: @context.my_type) if @context.my_type`

README.md
fgbulsoni: Nice catch :octocat:

libraries/resource_providers/image_streamer/api300/deployment_plan.rb
fgbulsoni: So, @tmiotto @jsmartt , thoughts about this? I'd guess the syntax would become easier to use, but if it seems unecessary just let me know.
Dropping the method would make it so that instead of this:
```ruby
@item['goldenImageURI'] = load_i3s_resource(:GoldenImage, @context.golden_image)
```
we'd use this:
```ruby          
@item['goldenImageURI'] = resource_named(:GoldenImage).find_by(@item.client, @context.golden_image).first['uri'] if @context.golden_image
```
Also, in this example, if `@context.golden_image`  contains a valid name, but no golden images exist in the appliance with that name, we'd hit:
```ruby
undefined method '[]' for nil:NilClass (NoMethodError)
```

tmiotto: Yep, it is easier to use the method. Keep it, but we need to return the entire resource instead only the uri.

libraries/resource_providers/image_streamer/api300/deployment_plan.rb
```diff
@@ -0,0 +1,33 @@
...
+      class DeploymentPlanProvider < ResourceProvider
+        def create_or_update
+          @item['goldenImageURI'] = load_resource(:GoldenImage, @context.golden_image)['uri']
+          @item['oeBuildPlanURI'] = load_resource(:BuildPlan, @context.build_plan)['uri']
+          super
+        end
+
+        def create_if_missing
+          @item['goldenImageURI'] = load_resource(:GoldenImage, @context.golden_image)['uri']
+          @item['oeBuildPlanURI'] = load_resource(:BuildPlan, @context.build_plan)['uri']
```
jsmartt: These params are listed as optional, so what happens here if the user doesn't set them? Won't this fail? I think we need to check to make sure it's set before setting these URIs

tmiotto: Makes sense, it should be `||=`

jsmartt: Right. But I think we need to add the `if @context.<attribute>` too. For example:

```ruby
@item['goldenImageURI'] ||= load_resource(:GoldenImage, @context.golden_image)['uri'] if @context.golden_image
```

tmiotto: The `load_resource` method is actually returning `nil` if the context property is not set.

jsmartt: Ah. I missed that. But it will still fail. `nil['uri']` doesn't work.

And I think it might be best to not have that method return nil if the name is not set. It should be up to the caller to call the method only if it needs to. If we can't use that special case here (because of the nil accessor error), I'm not sure where we'd be able to use it.

fgbulsoni: Yep, we were going one way trying to make the method as generic as possible, but it was quickly becoming harder to ensure all scenarios of failure. 
I've changed the method to be simple and accept the type & name, same as before, and return the uri. 
Should be useful for all the scenarios where we want the user to be able to pass in a name instead of uri, but is less generic now since it only looks for names.

libraries/resource_provider.rb
```diff
@@ -221,6 +221,16 @@ def get_diff(resource, desired_data)
...
     end
+
+    # Generic method to retrieve and return a resource from different resources using the type and name of the resource
+    # @param [String, Symbol] resource_class_type Type of resource to be retrieved. e.g., :GoldenImage, :FCNetwork.
+    # @param [String] resource_id Name for this resource. 'Resource1'
+    def load_resource_uri(resource_class_type, resource_id)
+      return unless resource_id
+      resource_instance = resource_named(resource_class_type).new(@item.client, name: resource_id)
+      raise "#{resource_class_type} resource with name '#{resource_id}' was not found in the appliance." unless resource_instance.retrieve!
+      resource_instance['uri']
```
jsmartt: This sure makes it easy to use, and I like it for this resource, but can I suggest just 1 more thing to make it more generic? What if it took an additional parameter, which describes what it should return. ie
```ruby
def load_resource(resource_class_type, resource_id, ret_attribute = nil)
# By default it returns the whole resource, but you can set ret_attribute to return a specific attribute
```
This way it can be used to return things other than just the uri

jsmartt (approved):
Cool. Looks good 👍 . We'll have to remember to comb though the rest of the code later and try to standardize using that cool new `load_resource` method.

libraries/resource_provider.rb
```diff
@@ -221,6 +221,18 @@ def get_diff(resource, desired_data)
     def recursive_diff(data, desired_data, str = '', indent = '')
       OneviewCookbook::Helper.recursive_diff(data, desired_data, str, indent)
     end
+
+    # Generic method to retrieve and return a resource from different resources using the type and name of the resource
+    # @param [String, Symbol] resource_class_type Type of resource to be retrieved. e.g., :GoldenImage, :FCNetwork.
+    # @param [String] resource_id Name for this resource. 'Resource1'
+    # @param [String, Symbol] ret_attribute Attribute of resource to be returned by this method. Use :resource for returning complete resource.
+    def load_resource(resource_class_type, resource_id, ret_attribute = :uri)
```
fgbulsoni, line 229: @tmiotto @jsmartt Also, just to point out that for now this is returning the `:uri` as default.
Will we want it that way, or do we rather have it return a different default?

tmiotto (approved):

libraries/resource_provider.rb
jsmartt: I'd probably vote for returning the whole resource by default, but we can easily change that later. I'm working on a different PR that may be able to use it, and I can refactor it then if need-be.

PR #207, opened 2017-03-06
Update README.md

jsmartt:
### Description
Updates the README to not say that an ImageStreamer appliance is required.

### Issues Resolved
(none)

### Check List
N/A

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,8 +11,8 @@ Chef cookbook that provides resources for managing HPE OneView.
 ## Requirements
 
  - Chef 12.0 or higher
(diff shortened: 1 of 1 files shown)


tmiotto:
I guess it would be good to also specify in which case you will need the Image Streamer appliance. Something like:

> Optional: HPE Synergy Image Streamer appliance (API version 300) (To be used only with image_streamer_<resource_name> resources)

jsmartt:
How about that?

tmiotto (approved):
Great :medal_sports:

PR #216, opened 2017-03-09
added guard

jsmartt:
### Description
Adds guard, which will watch for file changes and run tests automatically. In most cases it's smart enough to figure out which specific tests to run based on the file path, but the fixture recipes paths are too difficult to convert the test path. A different directory structure would help, but it's not worth changing for this purpose. It just runs a broader set of tests than it needs to.

You definitely don't have to use guard, but I've found it to be very helpful and a big time saver!

### Issues Resolved
(none)

### Check List
N/A


jsmartt:
Not sure what's affecting our coverage, but I'll take a look at it tomorrow

tmiotto:
@jsmartt 

I know what's affecting the coverage.

You moved the matchers_spec into the library folder.

The coverage report halts as soon the ChefSpec start running, so you need to run all the pure RSpec first, and then the ChefSpecs so the coverage report won't fail.

Gemfile
```diff
@@ -12,3 +12,15 @@ gem 'pry'
 gem 'rubocop', '= 0.40.0'
 gem 'simplecov'
 gem 'stove'
+
+begin
+  if Gem::Version.new(RUBY_VERSION) >= Gem::Version.new('2.2.5')
```
tmiotto: Guard officially supports Ruby '>= 2.2.6'

jsmartt:
Ah, thanks. I moved to so it could match based on the path, but I can move it back and set up a special rule for it. And I'll update the version too. Thanks :)

tmiotto (approved):
:taco:

PR #222, opened 2017-03-22
Cleaned up README

jsmartt:
### Description
Updates the README by removing unnecessary white-space and needless comments after each resource heading. It also updates some of the example code to be more accurate or clean, as well as adds a link to the example for each resource type heading.

The README is still huge, but this removes about 100  lines and adds more functionality.

### Issues Resolved
none

### Check List
N/A


tmiotto (approved):
:taco:

PR #223, opened 2017-03-23
oneview_storage_system should not try to update the name

### LINKED ISSUE #93: oneview_storage_system :update action shouldn't try to update name
### Scenario/Intent
Re-running the same recipe that added a storage system (with the `:add` action) always results in an update, regardless of whether or not it's needed. (On the name and password)

### Environment Details
 - **OneView Version:** 3.0

### Steps to Reproduce
Run any simple recipe with a oneview_storage_system resource with the `:add` action. The [example](https://github.com/HewlettPackard/oneview-chef/blob/master/examples/storage_system.rb#L25) will do.

### Expected Result
On subsequent runs, it shouldn't say it's going to update the name every time. From what I can tell, the name doesn't even appear to be configurable, right?
Also, it continues to try to update the password; I'm not sure if we can avoid that part, but what do you all think about only updating the password if the username has also changed or it's not in a good state (as defined by the state? status?)?

### Actual Result
It tries to update the StorageSystem name and password:

```
INFO: Updating oneview_storage_system '172.18.11.12'

    - Update oneview_storage_system '172.18.11.12'

      credentials:
        password: nil -> myPassword123
      name: ThreePAR7200-6955 -> 172.18.11.12
```

Note that it doesn't break anything or cause any errors, but it does throw off the resources updated count, etc.

jsmartt:
### Description
Disables the comparison of the storage_system's name since it can't be updated. Previously, it would try to update it every time the recipe was ran after the creation.

### Issues Resolved
Fixes #93

### Check List

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,8 +1,11 @@
-## 2.2.0
+## Unreleased (proposed: 2.2.0)
 Adds support to API300 HPE Synergy Image Streamer resources:
   - image_streamer_artifact_bundle
   - image_streamer_os_build_plan
 
+Bug Fixes:
+- [#93](https://github.com/HewlettPackard/oneview-chef/issues/93) oneview_storage_system should not try to update the name
+ 
 ## 2.1.0
 - [#162](https://github.com/HewlettPackard/oneview-chef/issues/162) Add server_profile_template property to oneview_server_profile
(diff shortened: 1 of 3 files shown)


tmiotto (approved):
:+1:

PR #224, opened 2017-03-23
Print diff before converge finished statement

### LINKED ISSUE #98: OneviewCookbook::Helper#get_diff with two no-diff hashes returns '\n'
### Scenario/Intent
When trying to compare two hashes that has no-diff ( i.e. for each key 'k' in 'a', a[k] == b[k] )  with the `OneviewCookbook::Helper#get_diff` the result returned is `'\n'`.

### Environment Details
 - **OneView Version:** N/A
 - **chef-client Version:** N/A
 - **platform:** N/A

### Steps to Reproduce
Use get_diff method with two hashes with no-diff like:
```ruby
a = { a: 'test', b: 'second' }
b = { a: 'test', c: 'other', b: 'second' }
get_diff(a, b)
```

### Expected Result
get_diff should return `''`

### Actual Result
get_diff returns `'\n'`

### LINKED ISSUE #145: Show diff on log statement before actual update
### Scenario/Intent
Resources currently print out the diff [here](https://github.com/HewlettPackard/oneview-chef/blob/master/libraries/resource_provider.rb#L70), after the change has been made to the resource, but I think it would be better to output it [here](https://github.com/HewlettPackard/oneview-chef/blob/master/libraries/resource_provider.rb#L65), before the change has been made. This is for a couple of reasons:
 - It's the resource diff that is **expected** to be updated, not the result. If things didn't actually get updated (e.g., StorageSystem name), it's misleading.
 - If the resource update fails the diff will never be printed out. Having it print out the diff before it tries the update will help figure out what you need to fix.

### Environment Details
 - **OneView Version:** All
 - **chef-client Version:** All
 - **platform:** All

### Steps to Reproduce
1. Create a resource (with Chef or directly on OneView)
2. Create a chef resource to try to update it, but give it invalid data that causes the API to return an error
3. Run that resource and look at the logs

### Expected Result
Show the resource diff (why it is going to try to do the update in the first place)

### Actual Result
No diff is shown; just the OneView error and/or SDK message

jsmartt:
### Description
This moves the diff statement to just before it tries to do the update of a resource, not after. This will be extremely helpful in debugging things, as often you don't know what it's trying to update if the update fails.

### Issues Resolved
Fixes #98
Fixes #145

### Check List

### DIFF
--- a/libraries/resource_providers/api300/synergy/fabric_provider.rb
+++ b/libraries/resource_providers/api300/synergy/fabric_provider.rb
@@ -24,7 +24,6 @@ module OneviewCookbook
             Chef::Log.info("#{resource_name} '#{name}' reserved Vlan range is up to date")
           else
             diff = get_diff(@item, options)
-            diff.insert(0, '. Diff:') unless diff.to_s.empty?
             Chef::Log.info "Setting #{resource_name} '#{name}' reserved Vlan range#{diff}"
             Chef::Log.debug "#{resource_name} '#{name}' Chef resource differs from OneView resource."
             Chef::Log.debug "Current state: #{JSON.pretty_generate(@item['reservedVlanRange'])}"
--- a/libraries/resource_providers/api300/synergy/fabric_provider.rb
+++ b/libraries/resource_providers/api300/synergy/fabric_provider.rb
@@ -23,12 +23,13 @@ module OneviewCookbook
           if @item.like? options
             Chef::Log.info("#{resource_name} '#{name}' reserved Vlan range is up to date")
           else
-            Chef::Log.info "Set #{resource_name} '#{name}' reserved Vlan range"
+            diff = get_diff(@item, options)
+            diff.insert(0, '. Diff:') unless diff.to_s.empty?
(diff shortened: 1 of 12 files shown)


libraries/resource_providers/api300/synergy/fabric_provider.rb
```diff
@@ -23,12 +23,13 @@ def set_reserved_vlan_range
           if @item.like? options
             Chef::Log.info("#{resource_name} '#{name}' reserved Vlan range is up to date")
           else
-            Chef::Log.info "Set #{resource_name} '#{name}' reserved Vlan range"
+            diff = get_diff(@item, options)
+            diff.insert(0, '. Diff:') unless diff.to_s.empty?
+            Chef::Log.info "Setting #{resource_name} '#{name}' reserved Vlan range#{diff}"
```
tmiotto: I see these 3 lines above repeated all over the code.

Could we change it to a method that receives the `actual`, the `compared`, and the `message`?

jsmartt: What if we just have the ResourceProvider method `#get_diff` do that for us?

tmiotto (approved):
Looks good! :medal_sports:

PR #226, opened 2017-03-24
Added Dockerfile & example instructions

jsmartt:
### Description
Adds a Dockerfile to make running examples easier for Docker users.

### Issues Resolved
none

### Check List
N/A

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,39 @@
+# See examples/README.md to see how to build and run this
+
+FROM ubuntu:16.04
+
+# Uncomment the next 2 lines if you're behind a proxy:
+ENV http_proxy="http://proxy.houston.hpecorp.net:8080"
+# ENV http_proxy="http://proxy.example.com:8080"
+ENV https_proxy=${http_proxy}
+RUN apt-get update && \
+    apt-get -yqq install \
+      curl \
+      unzip \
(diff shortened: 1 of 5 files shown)


Dockerfile
```diff
@@ -0,0 +1,38 @@
+# See examples/README.md to see how to build and run this
+
+FROM ubuntu:16.04
+
+# Uncomment the next 2 lines if you're behind a proxy:
+# ENV http_proxy="http://proxy.example.com:8080"
+# ENV https_proxy=${http_proxy}
```
tmiotto: I would just recommend adding a ENV no_proxy here too.

tmiotto (approved):
Great! :medal_sports: 

Optional - Add the NO_PROXY env var too in the Dockerfile

Dockerfile
jsmartt: I thought about that, but decided not to because I don't see how it'll ever be used. Is there a case you're thinking of where it would be?

fgbulsoni: Since we're usually connected to the appliances network through a VPN, when using vagrant for instance, we often use no_proxy to list exactly the appliance's IPs so those are accessible, otherwise just external sites become accessible with the http/https proxy. 
I'm assuming it's the same for docker.

PR #237, opened 2017-06-14
ID Pool for API 200/300 support

aalexmonteiro:
### Description
Implementation of the id pool for HPE OneView API300 support.

### Issues Resolved
[List any issues this PR will resolve]

### Check List


CHANGELOG.md
```diff
@@ -1,4 +1,7 @@
 ## 2.2.2 (Unreleased)
+Adds support to the following HPE OneView resources:
+  - Added oneview_id_pool resource
```
fgbulsoni: Since this release will add a new resource, it should likely be `2.3.0` instead of `2.2.2`

libraries/matchers.rb
```diff
@@ -29,6 +29,7 @@
     oneview_fc_network:                 standard_actions,
     oneview_fcoe_network:               standard_actions,
     oneview_firmware:                   [:add, :remove, :create_custom_spp],
+    oneview_id_pool:                    [:enable, :disable, :allocate_list, :allocate_count, :collect_ids, :validate],
```
fgbulsoni: A suggestion:

Perhaps instead of having many different actions for each thing, it actually makes sense to pack those together into an `update` action.
I'd imagine at least enable and disable could be inside an `update` action.

Also, `validate` does not seem to me like an action, it seems like something you'd call internally, perhaps before allocating a list. 

If this is similar to a get, it might be out of place. Does it cause any change on the appliance?

Are `allocate_list` and `collect_ids` opposites? As in, one allocates the IDs and the other deallocates them? Perhaps they could be inside an `update` or `set_ids` or some similar action?

fgbulsoni (changes requested):

aalexmonteiro:
@fgbulsoni , I made the changes, but in ```allocate_list``` and ```collect_ids```, I don't see these methods in a single method, because it is not a replacement of ids, I can add multiple ranges. On the UI of the OV, I see that.

README.md
```diff
@@ -810,6 +810,31 @@ oneview_user 'User1' do
...
+end
+
+oneview_id_pool 'IDPool1' do
+  client <my_client>
+  pool_type <pool_type> # String - The type of the pool. Values: (ipv4, vmac, vsn, vwwn)
+  id_list <id_list>     # Array<String> - The IDs list (or IDs separeted by comma)
+  action [:allocate_list, :collect_ids]
+end
+```
+
```
tmiotto: Please add also one block with the full description of the resource, i.e. with all the possible properties and actions listed

libraries/resource_providers/api200/id_pool_provider.rb
```diff
@@ -0,0 +1,49 @@
...
+        Chef::Log.info "Updating #{@resource_name} '#{@name}'"
+        @context.converge_by "Updating #{@resource_name} '#{@name}'" do
+          @item.update(enabled: @context.enabled)
+        end
+      end
+
+      def allocate_list
+        raise 'The IDs Pools list is not valid.' unless @item.validate_id_list(@context.pool_type, @context.id_list)
+        Chef::Log.info "Allocating the IDs #{@context.id_list} #{@resource_name} '#{@name}'"
+        @context.converge_by "Allocating the IDs #{@context.id_list} #{@resource_name} '#{@name}'" do
```
tmiotto: Change all the converge_by messages to past tense. They are only displayed after the actual execution occurs.

tmiotto (changes requested):

tmiotto:
I guess my comment on the review was not so clear. We want the full specification like this:
```ruby
 image_streamer_artifact_bundle 'ArtifactBundle1' do
   client <my_client>   # Hash or OneviewSDK::ImageStreamer::Client
   data <resource_data> # Hash
   deployment_plans <deployment_plan_names>
   golden_images <golden_image_names>
   os_build_plans <os_build_plan_names>
   plan_scripts <plan_script_names>
   new_name <artifact_bundle_name>          # String - The desired name for an existing artifact bundle - Optional
   file_path <local_file_path>              # String - File path for download & upload actions (Required for these actions)
   deployment_group <deployment_group_name> # String - Name of the deployment group on which to perform a backup. (Required for :backup_from_file action)
   timeout <timeout_value> # Integer - Time in seconds for the :backup_from_file action to timeout if it is not finished. - Optional
   action [:create_if_missing, :update_name, :delete, :download, :upload, :extract, :backup,
           :backup_from_file, :download_backup, :extract_backup]
 end
 ```

So it is easier for the user to see the full resource properties and uses.

But you can keep the written one you made too.

**Comment by aalexmonteiro** [comment]
@tmiotto , please see if I did it correctly now. :grin:

**Comment by tmiotto** [comment]
Yep, it is good, I would just reorder things:
1. Put on top the top-level block documentation;
2. Put the general description after that;
2. Put the action description right before **each** example:
 - **update:** Enable or disable an ID Pool. The Boolean property enabled is required.
```ruby
oneview_id_pool 'IDPool1' do
  client <my_client>    # Hash or OneviewSDK::ImageStreamer::Client
  pool_type <pool_type> # String - The type of the pool. Values: (ipv4, vmac, vsn, vwwn)
  enabled <enabled>     # [TrueClass, FalseClass] - The status of the pool
  action [:update]
end
```

And clarifying, you don't need to add an example for each action available **in the README**, that is why we link each resource to its corresponding **example**:
### [oneview_id_pool](examples/id_pool.rb)

tmiotto (approved):

fgbulsoni (approved):

PR #249, opened 2017-08-17
Upgraded gems and fixed consequences

### LINKED ISSUE #246: Upgrade oneview-sdk gem to version 5.0.0
### Scenario/Intent
Update oneview-sdk gem version to 5.0.0 that will provide support to HPE OneView API500 resources.

Update also the other gems if possible/applicable.

### Environment Details
 - **oneview Cookbook Version:** 2.3.0
 - **OneView Version:** 3.10

tmiotto:
### Description
Updates the gems, includig the oneview-sdk (5.0.0).

Also fixes all the consequences (mainly new rucobop style).

### Issues Resolved
#246 

### Check List
  - [ ] All tests pass (`$ rake test`).
- [ ] Changes documented in the [CHANGELOG](https://github.com/HewlettPackard/oneview-chef/blob/master/CHANGELOG.md).

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -2,10 +2,10 @@ source 'https://rubygems.org'
 
 ruby RUBY_VERSION # Needed to consider & solve for Ruby version requirements
 
-gem 'berkshelf', '~> 4.3.5'
+gem 'berkshelf'
 gem 'chef', '~> 12.0'
-gem 'chefspec', '~> 7.1'
-gem 'codeclimate-test-reporter', '~> 1.0.0'
+gem 'chefspec'
+gem 'codeclimate-test-reporter'
 gem 'foodcritic', '~> 7.1.0'
 gem 'oneview-sdk', '~> 5.0.0'
 gem 'pry'
--- a/Gemfile
+++ b/Gemfile
@@ -6,12 +6,12 @@ gem 'berkshelf', '~> 4.3.5'
 gem 'chef', '~> 12.0'
 gem 'chefspec', '~> 7.1'
 gem 'codeclimate-test-reporter', '~> 1.0.0'
-gem 'foodcritic', '~> 11.3'
+gem 'foodcritic', '~> 7.1.0'
 gem 'oneview-sdk', '~> 5.0.0'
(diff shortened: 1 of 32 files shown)


Gemfile
```diff
@@ -2,16 +2,16 @@ source 'https://rubygems.org'
...
 
-gem 'berkshelf'
+gem 'berkshelf', '~> 4.3.5'
 gem 'chef', '~> 12.0'
-gem 'chefspec'
+gem 'chefspec', '~> 7.1'
 gem 'codeclimate-test-reporter', '~> 1.0.0'
-gem 'foodcritic'
-gem 'oneview-sdk', '~> 4.5.1'
+gem 'foodcritic', '~> 11.3'
```
fgbulsoni: This is causing an error in the CI for Ruby 2.1:
```
    foodcritic (~> 11.3) was resolved to 11.3.0, which depends on
      ruby (>= 2.2.2)
```
Should likely add it in a condition, similar to the guard gem, to use this version in Ruby 2.2+ and use the latest possible/available for Ruby 2.1.

fgbulsoni (changes requested):

Gemfile
tmiotto: Actually it is not the best approach.

I'll lower the version binding, it is not safe to leave it causing issues as new versions appears in the new rubies. (same as rubocop rules)

fgbulsoni: > I'll lower the version binding, it is not safe to leave it causing issues as new versions appears in the new rubies

Right, but that version is only causing issues in Ruby 2.1. 
Should we really stick to a lower version of that gem on all rubies, just for ruby 2.1 compatibility? 
Also, if it actually dropped ruby 2.1 support, that'd mean we'd be stuck on that version permanently.

tmiotto: > Should we really stick to a lower version of that gem on all rubies, just for ruby 2.1 compatibility?

Yes, the rules may change from one version to another, and we do not want to keep two different rule files and so on.

fgbulsoni (approved):

PR #279, opened 2017-08-25
API 500 Support for Managed SAN

### LINKED ISSUE #275: Support for Managed SAN - API500
### Scenario/Intent
Add support to Managed SAN on the API500.

### Environment Details
 - **OneView Version:** 3.10
 - **platform:** Any

### Steps to Reproduce
Try to use the resource Managed SAN on the API 500.

### Expected Result

### Actual Result
This resource does not support API 500.

aalexmonteiro:
### Description
Adding Managed SAN for HPE OneView API500 support.

### Issues Resolved
#275 

### Check List

### DIFF
--- a/libraries/resource_providers/api200/managed_san_provider.rb
+++ b/libraries/resource_providers/api200/managed_san_provider.rb
@@ -33,16 +33,11 @@ module OneviewCookbook
       end
 
       def set_public_attributes
-        temp = Marshal.load(Marshal.dump(@item['publicAttributes']))
+        temp_attributes = Marshal.load(Marshal.dump(@item['publicAttributes']))
         @item.retrieve! || raise("#{@resource_name} '#{@name}' not found!")
         # compares two hashes
-        results = []
-        temp_attributes = temp.sort_by { |element| element['name'] }
-        item_attributes = @item['publicAttributes'].sort_by { |element| element['name'] }
-        temp_attributes = temp_attributes.each_with_index do |element, index|
-          results << element.all? { |k, v| v == item_attributes[index][k] }
-        end
-        return Chef::Log.info "#{@resource_name} '#{@name}' public attributes are up to date" if results.all?
+        results = (temp_attributes - @item['publicAttributes']) + (@item['publicAttributes'] - temp_attributes)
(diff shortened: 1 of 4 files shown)


libraries/resource_providers/api200/managed_san_provider.rb
```diff
@@ -35,16 +35,11 @@ def set_policy
...
         @item.retrieve! || raise("#{@resource_name} '#{@name}' not found!")
         # compares two hashes
-        results = []
-        temp_attributes = temp.sort_by { |element| element['name'] }
-        item_attributes = @item['publicAttributes'].sort_by { |element| element['name'] }
-        temp_attributes = temp_attributes.each_with_index do |element, index|
-          results << element.all? { |k, v| v == item_attributes[index][k] }
-        end
-        return Chef::Log.info "#{@resource_name} '#{@name}' public attributes are up to date" if results.all?
+        results = (temp_attributes - @item['publicAttributes']) + (@item['publicAttributes'] - temp_attributes)
```
fgbulsoni, line 39: It appears there's quite some effort put into sorting the elements before, and now that is gone. Are we *certain* that is not needed?

aalexmonteiro: I made the change due to two occurrences.

1- An error occurs when, after recovering san managed and the ```publicAttributes``` attribute is empty, breaking inside the loop.

2 - I understand that the filter gets only the attribute that should be updated. But in the Oneview documentation it is quoted:  

```Updating a specific Managed SAN's publicAttributes. The contents of the PUT publicAttributes will replace the existing list so leaving out a public attribute from the list will effectively delete it.```

So if there is an array with two hashes of attributes, the same one that sends the two hashes, but only one of them has changed, only the hash changes are sent in the load and Oneview will discard the attributes of the another hash.

Correct me if I'm wrong, please!

fgbulsoni: 2 does make sense, specially if considering the name 'set_public_attributes' which implies that the public attributes will be exactly as specified.
Ok, I've no issues with the change.

fgbulsoni:
>> Fixing merge conflicts
--

Care with these needless commits

fgbulsoni (approved):

PR #293, opened 2017-08-29
API 500 Support for Logical Interconnect Group

### LINKED ISSUE #278: Support for Logical Interconnect Group - API500
### Scenario/Intent
Add support to Logical Interconnect Group on the API500.

### Environment Details
 - **OneView Version:** 3.10
 - **OneView API Version:** 500
 - **platform:** Any

### Steps to Reproduce
Try to use the resource Logical Interconnect Group on the API 500.

### Expected Result

### Actual Result
This resource does not support API 500.

ricardogpsf:
### Description
API 500 Support for Logical Interconnect Group.

### Issues Resolved
#278   

### Check List


examples/logical_interconnect_group.rb
```diff
@@ -12,12 +12,15 @@
 # NOTE: This recipe requires:
 # Ethernet Networks: EthernetNetwork1, EthernetNetwork2
 # FC Network: FCNetwork1
+# Scopes: Scope1, Scope2
 
+# NOTE: The api_version client should be greater than 200 if you run the examples using Scopes
```
fgbulsoni: I'd swap `greater than 200` for `(...) 300 or greater (...)`. Since 200 is not supported, it probably makes sense to not even mention it.

tmiotto (approved):
🍰 ✨

fgbulsoni (approved):
The PR overall looks good, but one thing I'm wondering:
Most of the resources that gained the add/remove/replace scopes will have tests that are basically the same as these, right? 
Perhaps it is worth creating a `shared` base for these, which will only be included by the resources and then each one could pass in its type through a `let` for instance?
This is not a *must*, but I do imagine that the resources which will gain those actions will have tests that are pretty similar

ricardogpsf:
> The PR overall looks good, but one thing I'm wondering:
Most of the resources that gained the add/remove/replace scopes will have tests that are basically the same as these, right?
Perhaps it is worth creating a shared base for these, which will only be included by the resources and then each one could pass in its type through a let for instance?
This is not a must, but I do imagine that the resources which will gain those actions will have tests that are pretty similar

@fgbulsoni yes, it is good idea. =) 
We can do that in another PR, what do you think?

tmiotto:
> We can do that in another PR, what do you think?

Issue it, PR it, We review it, merge it, Done.

fgbulsoni:
> Issue it, PR it, We review it, merge it, Done.

Also, doing this sooner than later will mean less effort spent repeating tests. Doing it later will just mean reworking the tests for not a whole lot of gain.

ricardogpsf:
Issue #306 registered.

PR #308, opened 2017-09-05
Server Profile API500 Support

fgbulsoni:
### Description
Allowed FCoE Networks on connections and validated against API500
- New property fcoe_network_connections added
- Validated against API500
- Docs updated

### Check List
  - [X] All tests pass (`$ rake test`).
- [X] New functionality has been documented in the README if applicable.
  - [X] New functionality has been thoroughly documented in [examples](https://github.com/HewlettPackard/oneview-chef/tree/master/examples/) (please include helpful comments).
- [X] Changes documented in the [CHANGELOG](https://github.com/HewlettPackard/oneview-chef/blob/master/CHANGELOG.md).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ Adds API 500 support to the following HPE OneView resources:
   - oneview_sas_logical_interconnect_group
   - oneview_scope
   - oneview_server_hardware_type
+  - oneview_server_profile
   - oneview_storage_system
   - oneview_unmanaged_device
   - oneview_user
--- a/README.md
+++ b/README.md
@@ -767,6 +767,7 @@ oneview_server_profile 'ServerProfile1' do
   firmware_driver <firmware_driver_name>
   ethernet_network_connections <ethernet_network_connections_data>
   fc_network_connections <fc_network_connections_data>
+  fcoe_network_connections <fcoe_network_connections_data>
   network_set_connections <network_set_connections_data>
   os_deployment_plan <os_deployment_plan_name>
   action [:create, :create_if_missing, :delete]
--- a/examples/server_profile.rb
+++ b/examples/server_profile.rb
@@ -71,6 +71,13 @@ oneview_server_profile 'ServerProfile3' do
       'portId' => 'Auto'
     }
   )
+  fcoe_network_connections(
+    'fcoe1' => {
+      'name' => 'c1',
+      'functionType' => 'FibreChannel',
+      'portId' => 'Auto'
+    }
+  )
   action :create_if_missing
 end
 
--- a/examples/server_profile.rb
+++ b/examples/server_profile.rb
@@ -15,11 +15,16 @@ my_client = {
   password: ENV['ONEVIEWSDK_PASSWORD']
 }
 
+my_server_profile_template = 'spt1'
+my_server_hardware_type = 'SY 480 Gen9 2'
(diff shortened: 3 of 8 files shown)


fgbulsoni:
> 1 failing and 2 successful checks

@tmiotto @ricardogpsf @aalexmonteiro  I'm seeing a lot of failures on the PR merge, mostly mentioning this:

```
An error occurred while loading ./spec/unit/libraries/oneview_resource_base_spec.rb.
Failure/Error: Dir[File.expand_path('../libraries/*.rb', File.dirname(__FILE__))].each { |file| require file }
NameError:
  uninitialized constant OneviewCookbook::API200::ServerProfileTemplateProvider::ServerProfileProviderHelpers
``` 

And that should not have been touched by this PR in any way. 

Ideas as to why those errors are happening?

ricardogpsf:
@fgbulsoni I guess the problem is occuring because the server_profile_template_provider.rb is not requiring the file where ServerProfileProviderHelpers is. Depending of way that file are imported when we use `... '../libraries/*.rb' ...` the ServerProfileProviderHelpers can be imported after the server_profile_template_provider.rb, generating the problem. 
I think that we can solve this problem sorting all files imported by name:
`Dir[File.expand_path('../libraries/*.rb', File.dirname(__FILE__))].sort.each { |file| require file }`
or
importing individually the necessary files.

What do you think?

aalexmonteiro:
A question:

In the Ruby SDK there is a method for creating a server profile with a volume attachment ```add_volume_attachment```.

Would it be nice to have a ```volume_attachment``` property?

fgbulsoni:
> Dir[File.expand_path('../libraries/*.rb', File.dirname(__FILE__))].sort.each { |file| require file }

While that could work, it'd be more of a 'masking the problem' than solving it per say, as that would just accentuate the issue in case of file started with `a` that needs to import from one started with `b`, and it'd likely have some bad effect in performance as well.

I'll try adding a `require_relative` in the spt provider to see if that solves this issue for now, and eventually we'll need to come up with better alternatives I guess.

ricardogpsf:
The idea with `Dir[File.expand_path('../libraries/*.rb', File.dirname(FILE))].sort.each { |file| require file }` is ensure that problem will not be inconstant, but will be expected to throw an error when should throw an error. 
Because currently the problem occurs in some environments and not occurs in another environments, I believe (I do not searched very much about it) that ruby does not ensure the order of files when we use `Dir[File.expand_path('../libraries/*.rb', File.dirname(FILE))]`, so the order of this files can be diferently of expected.

I agree that the correct way looks using the `require_relative`, but I guess is good sorting the files imported because can avoid the developer forget the `require_relative` and the error not occur, when the imported files are sorted we are forcing the throwing of an error.

tmiotto:
@ricardogpsf  If the developer forgets to put something that he is intended to put, it is a bug. The requires must not be forgotten.

tmiotto (commented):
Missing Unit tests

tmiotto (commented):
I know it was not added in this PR but you could also fix the example? I'm seeing json notation amidst the ruby hashes there

ricardogpsf:
> @ricardogpsf If the developer forgets to put something that he is intended to put, it is a bug. The requires must not be forgotten.

@tmiotto ok, no problem. =)

tmiotto (approved):
🦄✨

PR #324, opened 2017-09-13
API500 Support to Power Device

ricardogpsf:
### Description
- API500 Support to Power Device
- Added new actions for oneview_power_device: `refresh`, `set_power_state`, `set_uid_state`.

### Environment Details
 - **OneView Version:** 3.10
 - **platform:** Any

### Steps to Reproduce
Try to use the oneview_power_deviceon the API 500.

### Expected Result

### Check List

### DIFF
--- a/libraries/resource_providers/api200/power_device_provider.rb
+++ b/libraries/resource_providers/api200/power_device_provider.rb
@@ -23,13 +23,52 @@ module OneviewCookbook
         end
       end
 
+      def load_power_device
+        return @item if @item.retrieve!
+        power_devices_list = resource_named(:PowerDevice).get_ipdu_devices(@item.client, @name)
+        power_devices_list.first
+      end
+
       def remove
         # First try to remove by name, if it does not work it consider the power device is an iPDU
         return true if super
-        power_devices_list = resource_named(:PowerDevice).get_ipdu_devices(@item.client, @name)
-        return false if power_devices_list.empty?
-        @item = power_devices_list.first
+        item = load_power_device
+        return false if item.nil?
+        @item = item
         super
+      end
+
+      def refresh
+        @item = load_power_device || raise("#{@resource_name} '#{@name}' not found!")
+        raise "Unspecified property: 'refresh_options'. Please set it before attempting this action." unless @new_resource.refresh_options
+        refresh_ready = ['RefreshFailed', 'NotRefreshing', ''].include? @item['refreshState']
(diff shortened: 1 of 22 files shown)


libraries/resource_providers/api200/power_device_provider.rb
```diff
@@ -23,14 +23,53 @@ def discover
...
-        power_devices_list = resource_named(:PowerDevice).get_ipdu_devices(@item.client, @name)
-        return false if power_devices_list.empty?
-        @item = power_devices_list.first
+        item = load_power_device
+        return false if item.nil?
+        @item = item
         super
       end
+
+      def refresh
```
tmiotto, line 41: You could use the refresh Mixin, don't you?

spec/unit/resources/power_device/refresh_spec.rb
```diff
@@ -0,0 +1,46 @@
+require_relative './../../../spec_helper'
+
+describe 'oneview_test::power_device_refresh' do
+  let(:resource_name) { 'power_device' }
+  include_context 'chef context'
+
+  let(:target_class) { OneviewSDK::API200::PowerDevice }
+
+  context 'when Oneview resource can be retrieved' do
```
tmiotto, line 9: I guess this could be added to a shared example. Many resources use this kind of refresh.

libraries/resource_providers/api200/power_device_provider.rb
ricardogpsf: I thought about this, but this `refresh` call the method `set_refresh_state` that use a hash of options as argument (the string of state is into the options), and it is the unique case of this method set_refresh_state with this parameter. =(
The other reason is when we retrieve the resource item, we made that in different way, because the item can be a IPDU. 
So for this reasons, I thought be better create a new method into power_device instead of change the `refresh` to a particular case.

I thought too using the property `refresh_state` of power_device resource to receive a hash with options, so I could call the refresh of mixin (adjusting it to call some method to retrieve the resource item, because in case of "powser device" can be IPDU and it is retrieved in differente way), but I imaginated that would be wrong semantic of property `refresh_state`, once that would receive the other options like username, password.

What do you think?

spec/unit/resources/power_device/refresh_spec.rb
ricardogpsf: Same reason that previous comment.

libraries/resource_providers/api200/power_device_provider.rb
tmiotto: Ok, no problem

tmiotto (approved):
:potato: ✨

PR #325, opened 2017-09-14
API 500 Support for Logical Interconnect

aalexmonteiro:
### Description
Adding Logical Interconnect for HPE OneView API500 support.

### Issues Resolved
N/A

### Check List

### DIFF
--- a/README.md
+++ b/README.md
@@ -283,16 +283,17 @@ Note: By default it performs the action `:none`.
 oneview_logical_interconnect 'LogicalInterconnect1' do
   client <my_client>
   data <resource_data>
-  firmware <firmware_name>           # String: Optional for actions like :<action>_firwmare (can be replaced by data attribute 'sppName')
-  firmware_data <firmware_data>      # Hash: Optional for actions like :<action>_firwmare
-  internal_networks <networks_names> # Array: Optional for :update_internal_networks
-  trap_destinations <trap_options>   # Hash: Optional for :update_snmp_configuration
-  enclosure <enclosure_name>         # String: Required for :add_interconnect and :remove_interconnect
-  bay_number <bay>                   # Integer: Required for :add_interconnect and :remove_interconnect
-  operation <op>                     # String. Used in patch action only. e.g., 'add'
-  path <path>                        # String. Used in patch option only. e.g., '/scopeUris/-'
-  value <val>                        # String. Used in patch option only. e.g., 'scope uri'
-  scopes <scope_names>               # Array - Optional. Array of scope names. Used in add_to_scopes, remove_from_scopes or replace_scopes options only. e.g., ['Scope1', 'Scope2']
+  firmware <firmware_name>            # String: Optional for actions like :<action>_firwmare (can be replaced by data attribute 'sppName')
+  firmware_data <firmware_data>       # Hash: Optional for actions like :<action>_firwmare
+  internal_networks <networks_names>  # Array: Optional for :update_internal_networks
+  trap_destinations <trap_options>    # Hash: Optional for :update_snmp_configuration
+  enclosure <enclosure_name>          # String: Required for :add_interconnect and :remove_interconnect
+  bay_number <bay>                    # Integer: Required for :add_interconnect and :remove_interconnect
+  port_monitor <port_monitor_options> # Hash: Optional for :update_port_monitor
+  operation <op>                      # String. Used in patch action only. e.g., 'add'
+  path <path>                         # String. Used in patch option only. e.g., '/scopeUris/-'
+  value <val>                         # String. Used in patch option only. e.g., 'scope uri'
+  scopes <scope_names>                # Array - Optional. Array of scope names. Used in add_to_scopes, remove_from_scopes or replace_scopes options only. e.g., ['Scope1', 'Scope2']
   action [:none, :add_interconnect, :remove_interconnect, :update_internal_networks,
           :update_settings,:update_ethernet_settings, :update_port_monitor, :update_qos_configuration,
           :update_telemetry_configuration, :update_snmp_configuration, :update_firmware, :stage_firmware,
--- a/README.md
+++ b/README.md
@@ -278,7 +278,8 @@ end
 Performs actions on logical interconnect and associated interconnects.
 
 Note: By default it performs the action `:none`.
-Note 2: In the `:update_port_monitor` action, if the same field is informed inside both `data` and `port_monitor`, the value of that field inside data will supersede the value inside `port_monitor`.
+
+Note: In the `:update_port_monitor` action, if the same field is informed inside both `data` and `port_monitor`, the value of that field inside `data` will supersede the value inside `port_monitor`.
 
 ```ruby
 oneview_logical_interconnect 'LogicalInterconnect1' do
--- a/README.md
+++ b/README.md
@@ -278,7 +278,7 @@ end
 Performs actions on logical interconnect and associated interconnects.
 
 Note: By default it performs the action `:none`.
-Note 2: In :update_port_monitor action, if the data is informed, this will always precede the port_monitor property.
+Note 2: In the `:update_port_monitor` action, if the same field is informed inside both `data` and `port_monitor`, the value of that field inside data will supersede the value inside `port_monitor`.
 
 ```ruby
 oneview_logical_interconnect 'LogicalInterconnect1' do
--- a/README.md
+++ b/README.md
@@ -278,6 +278,7 @@ end
 Performs actions on logical interconnect and associated interconnects.
 
 Note: By default it performs the action `:none`.
+Note 2: In :update_port_monitor action, if the data is informed, this will always precede the port_monitor property.
 
 ```ruby
 oneview_logical_interconnect 'LogicalInterconnect1' do
--- a/README.md
+++ b/README.md
@@ -289,10 +289,15 @@ oneview_logical_interconnect 'LogicalInterconnect1' do
   trap_destinations <trap_options>   # Hash: Optional for :update_snmp_configuration
   enclosure <enclosure_name>         # String: Required for :add_interconnect and :remove_interconnect
   bay_number <bay>                   # Integer: Required for :add_interconnect and :remove_interconnect
+  operation <op>                     # String. Used in patch action only. e.g., 'add'
+  path <path>                        # String. Used in patch option only. e.g., '/scopeUris/-'
+  value <val>                        # String. Used in patch option only. e.g., 'scope uri'
+  scopes <scope_names>               # Array - Optional. Array of scope names. Used in add_to_scopes, remove_from_scopes or replace_scopes options only. e.g., ['Scope1', 'Scope2']
   action [:none, :add_interconnect, :remove_interconnect, :update_internal_networks,
           :update_settings,:update_ethernet_settings, :update_port_monitor, :update_qos_configuration,
           :update_telemetry_configuration, :update_snmp_configuration, :update_firmware, :stage_firmware,
-          :activate_firmware, :update_from_group, :reapply_configuration]
+          :activate_firmware, :update_from_group, :reapply_configuration, :patch, :add_to_scopes,
+          :remove_from_scopes, :replace_scopes]
 end
 ```
 
--- a/examples/logical_interconnect.rb
+++ b/examples/logical_interconnect.rb
@@ -63,21 +63,31 @@ end
 # Activate the port monitor service
 oneview_logical_interconnect 'Encl1-LogicalInterconnectGroup1' do
   client my_client
-  data(
-    portMonitor: {
-      analyzerPort: {
-        portUri: '/rest/interconnects/c3498956-ece5-4abf-9e7f-793cee92acca/ports/c3498956-ece5-4abf-9e7f-793cee92acca:1',
-        portMonitorConfigInfo: 'AnalyzerPort'
-      },
-      enablePortMonitor: true,
-      type: 'port-monitor',
-      monitoredPorts: [
-        {
-          portUri: '/rest/interconnects/c3498956-ece5-4abf-9e7f-793cee92acca/ports/c3498956-ece5-4abf-9e7f-793cee92acca:d3',
-          portMonitorConfigInfo: 'MonitoredBoth'
-        }
-      ]
-    }
+  port_monitor(
+    analyzerPort: {
+      port_name: 'Q1.3',
+      portMonitorConfigInfo: 'AnalyzerPort'
+    },
+    enablePortMonitor: true,
+    type: 'port-monitor',
+    monitoredPorts: [
+      {
+        port_name: 'd1',
+        portMonitorConfigInfo: 'MonitoredBoth'
+      }
+    ]
+  )
+  action :update_port_monitor
+end
+
+# Disable the port monitor service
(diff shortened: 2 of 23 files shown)


libraries/resource_providers/api200/logical_interconnect_provider.rb
```diff
@@ -38,7 +38,7 @@ def interconnect_handler(present_block, absent_block)
       # Recursive helper method to set hash values
       def recursive_set(actual, updates)
         updates.each_pair do |k, v|
-          if v.respond_to?(:each_pair)
+          if v.respond_to?(:each_pair) && actual[k].respond_to?(:each_pair)
```
aalexmonteiro, line 41: I've added ```actual[k].respond_to?(:each_pair)``` in the if clause because if to be passed a hash with attributes and these attributes is not present in the actual hash, an undefined method occurs: ```undefined method [] = 'for nil: NilClass```.

examples/logical_interconnect.rb
```diff
@@ -99,18 +112,18 @@
...
   )
   trap_destinations(
-    '172.18.6.16' => {
-      'trapFormat' => 'SNMPv1',
-      'communityString' => 'public',
-      'severities' => ['Critical', 'Major', 'Unknown'],
-      'vcmTraps' => ['Legacy'],
-      'enetTraps' => ['PortStatus', 'PortThresholds'],
-      'fcTraps' => ['Other']
+    '172.18.6.16': {
```
tmiotto: I don't thik it will work

examples/logical_interconnect.rb
```diff
@@ -62,8 +64,19 @@
 oneview_logical_interconnect 'Encl1-LogicalInterconnectGroup1' do
   client my_client
   data(
-    'portMonitor' => {
-      'enabled' => true
+    portMonitor: {
+      analyzerPort: {
+        portUri: '/rest/interconnects/c3498956-ece5-4abf-9e7f-793cee92acca/ports/c3498956-ece5-4abf-9e7f-793cee92acca:1',
```
ricardogpsf: Just a question, the value of portUri is variable or is previsible? How will client set this value at runtime of recipes?

examples/logical_interconnect.rb
aalexmonteiro: I also had this doubt, but I tested it and Ruby converts that string into a symbol and that worked fine. But if this is strange, I can modify it.

examples/logical_interconnect.rb
aalexmonteiro: It's ok! I'll implement a way to get the uri by the port name. What do you think?

ricardogpsf: Nice! =)

examples/logical_interconnect.rb
tmiotto: Chnage to string hash syntax `'x' => y`

libraries/oneview_helper.rb
```diff
@@ -168,7 +168,11 @@ def self.convert_keys(info, conversion_method)
       return nil unless info
       support = {}
       info.each do |k, v|
-        con = convert_keys(v, conversion_method) if v && v.class == Hash
+        if v && v.class == Hash
```
ricardogpsf: just an observation: It is not necessary verify the `if v` because if `v` is a `Hash`, it is not `nil`.

ricardogpsf (approved):

libraries/resource_providers/api200/logical_interconnect_provider.rb
```diff
@@ -82,6 +82,26 @@ def firmware_handler(action)
...
+        port = @item.get_unassigned_uplink_ports_for_port_monitor.select { |k| k['portName'] == ports['analyzerPort']['port_name'] }.first
+        raise("Port '#{ports['analyzerPort']['port_name']}' was not found or is already being monitored!") unless port
+        load_downlink_ports(ports['monitoredPorts'], port['interconnectName'])
+        ports['analyzerPort']['portUri'] = port['uri']
+        ports['analyzerPort'].delete('port_name')
+        ports['monitoredPorts'].each { |k| k.delete('port_name') }
+      end
+
+      def load_downlink_ports(monitored_ports, interconnect_name)
+        interconnect = resource_named(:Interconnect).find_by(@item.client, name: interconnect_name).first
```
fgbulsoni:
```ruby
interconnect = load_resource(:Interconnect, interconnect_name)
```

libraries/resource_providers/api200/logical_interconnect_provider.rb
```diff
@@ -129,6 +149,10 @@ def update_ethernet_settings
       end
 
       def update_port_monitor
+        @item.retrieve! || raise("#{@resource_name} '#{@name}' not found!")
+        ports = convert_keys(@new_resource.port_monitor, :to_s)
+        load_port(ports) if ports['enablePortMonitor']
+        @item['portMonitor'] = ports
```
fgbulsoni: Why not include this into `load_port` ?
Also, what if the user specifies a `data['portMonitor']` ? It is currently being overwritten by even `{}`, which is the default for `port_monitor`, right?

libraries/resource_providers/api200/logical_interconnect_provider.rb
```diff
@@ -129,6 +149,10 @@ def update_ethernet_settings
       end
 
       def update_port_monitor
+        @item.retrieve! || raise("#{@resource_name} '#{@name}' not found!")
+        ports = convert_keys(@new_resource.port_monitor, :to_s)
```
fgbulsoni: likely better included inside `load_port`

libraries/resource_providers/api200/logical_interconnect_provider.rb
```diff
@@ -82,6 +82,26 @@ def firmware_handler(action)
         end
       end
 
+      def load_port(ports)
```
fgbulsoni: shouldn't it be `load_ports`?

fgbulsoni (changes requested):

libraries/resource_providers/api200/logical_interconnect_provider.rb
aalexmonteiro: Best not, because ``` loadPorts``` will only be executed if the attribute ```enablePortMonitor``` is true. I don't think "ports" is a good name. I think I'd better to change to  "port_monitor". What do you think?

libraries/resource_providers/api200/logical_interconnect_provider.rb
aalexmonteiro: I understood. I'll change to prevent this from happening.

libraries/resource_providers/api200/logical_interconnect_provider.rb
fgbulsoni: I believe running `convert_keys` to be more costly outside than inside the `load_port` method, as you'd just need to change:
```ruby
load_port(ports) if ports['enablePortMonitor']
```
to
```ruby
load_port(ports) if @new_resource.port_monitor.any?
```
or something similar and that method wouldn't even be called if port_monitor is not defined.

fgbulsoni: As for the name of the variable, both `ports` and `port_monitor` are somewhat abstract to me.
Perhaps `port_monitor_attributes`? Or something similar. 
Just suggestions, not a particular issue to me with the naming at least.

libraries/resource_providers/api200/logical_interconnect_provider.rb
```diff
@@ -82,17 +82,22 @@ def firmware_handler(action)
...
 
-      def load_port(ports)
-        port = @item.get_unassigned_uplink_ports_for_port_monitor.select { |k| k['portName'] == ports['analyzerPort']['port_name'] }.first
-        raise("Port '#{ports['analyzerPort']['port_name']}' was not found or is already being monitored!") unless port
-        load_downlink_ports(ports['monitoredPorts'], port['interconnectName'])
-        ports['analyzerPort']['portUri'] = port['uri']
-        ports['analyzerPort'].delete('port_name')
-        ports['monitoredPorts'].each { |k| k.delete('port_name') }
+      def load_ports
+        @item.retrieve! || raise("#{@resource_name} '#{@name}' not found!")
```
fgbulsoni: This is likely better off in the `update_port_monitor` method, before calling `load_ports`, as you want to fail asap if the LI doesn't exist.

libraries/resource_providers/api200/logical_interconnect_provider.rb
```diff
@@ -82,17 +82,22 @@ def firmware_handler(action)
...
+        port_monitor_attributes = convert_keys(@new_resource.port_monitor, :to_s)
+        if port_monitor_attributes['enablePortMonitor']
+          port = @item.get_unassigned_uplink_ports_for_port_monitor.select { |k| k['portName'] == port_monitor_attributes['analyzerPort']['port_name'] }.first
+          raise("Port '#{port_monitor_attributes['analyzerPort']['port_name']}' was not found or is already being monitored!") unless port
+          load_downlink_ports(port_monitor_attributes['monitoredPorts'], port['interconnectName'])
+          port_monitor_attributes['analyzerPort']['portUri'] = port['uri']
+          port_monitor_attributes['analyzerPort'].delete('port_name')
+          port_monitor_attributes['monitoredPorts'].each { |k| k.delete('port_name') }
+        end
+        @item['portMonitor'] = port_monitor_attributes
```
fgbulsoni: This is still overwriting the `@item['portMonitor']`, why not merge the two here?
That way:
1- If a `data['portMonitor']` is specified, it should be used
2- If a `port_monitor` parameter is specified, it should be used
3- If **both** are specified, end result is a merge of `data['portMonitor']` + `port_monitor_attributes`

fgbulsoni (changes requested):

libraries/resource_providers/api200/logical_interconnect_provider.rb
aalexmonteiro: I thought of doing it that way. Then I realized that I would have to manipulate another object, like ```temp = key ? { key.to_s => Marshal.load(Marshal.dump(@item[key])) } : Marshal.load(Marshal.dump(@item.data))``` and it would practically be difficult to reuse the ```update_handler``` method.

fgbulsoni: Not sure I understand. 

Why would you need to do all that to run `@item.retrieve! || raise("#{@resource_name} '#{@name}' not found!")` in `update_port_monitor` instead of here?

If you just move this from [line 86](https://github.com/HewlettPackard/oneview-chef/pull/325/files/cb5f7cc7d99d73a925bedd963b466960cf9e7e00..c59a4c4e2bf7d9b21c1582ff5394903025db65e6#diff-4b902556c979d8bc0791ddf2cac48843R86) to [line 157](https://github.com/HewlettPackard/oneview-chef/pull/325/files/cb5f7cc7d99d73a925bedd963b466960cf9e7e00..c59a4c4e2bf7d9b21c1582ff5394903025db65e6#diff-4b902556c979d8bc0791ddf2cac48843R157) it should work just fine unless I'm missing something.

aalexmonteiro: Sorry @fgbulsoni , I was not clear. When I run an ```@item.retrieve!```. the data is replaced. Being necessary to store in another variable, as is done in the ```update_handler```.

fgbulsoni: Ok, so, both of that load/unload and the retrieve likely belong in the `update_port_monitor` method, I'm guessing in that order. Right?

aalexmonteiro: Yeah! I divided the algorithm into more than one method to split the tasks.

libraries/resource_providers/api200/logical_interconnect_provider.rb
```diff
@@ -129,6 +156,10 @@ def update_ethernet_settings
       end
 
       def update_port_monitor
+        temp = Marshal.load(Marshal.dump(@item.data))
```
fgbulsoni: temp is not a good var name. Something like `new_data` probably would make sense

libraries/resource_providers/api200/logical_interconnect_provider.rb
```diff
@@ -82,6 +82,33 @@ def firmware_handler(action)
         end
       end
 
+      def load_ports(item_to_update)
+        port_monitor_attributes = convert_keys(@new_resource.port_monitor, :to_s)
+        enable_port_monitor = item_to_update['portMonitor']['enablePortMonitor'] if item_to_update['portMonitor'] && item_to_update['portMonitor']['enablePortMonitor']
```
fgbulsoni: the `&& item_to_update['portMonitor']['enablePortMonitor']` is likely not necessary and should be dropped.
Theis code:
```ruby
enable_port_monitor = item_to_update['portMonitor']['enablePortMonitor'] if item_to_update['portMonitor']
```
Should return `nil` if `'enablePortMonitor'` does not exist, and it will be overwritten bellow through the `||=`

libraries/resource_providers/api200/logical_interconnect_provider.rb
```diff
@@ -82,6 +82,33 @@ def firmware_handler(action)
...
+        enable_port_monitor ||= port_monitor_attributes['enablePortMonitor']
```
fgbulsoni: Also, the way this is setup make it so `data['portMonitor']['enablePortMonitor']` has precedence over `@new_resource.port_monitor['enablePortMonitor']`. 
Is this the expected behavior? If it is, we want to be sure to inform that `data > new_resource` if both are declared

examples/logical_interconnect.rb
```diff
@@ -61,9 +63,66 @@
 # Activate the port monitor service
 oneview_logical_interconnect 'Encl1-LogicalInterconnectGroup1' do
   client my_client
+  port_monitor(
```
fgbulsoni, line 66: Also, IF we want to be merging and using both `data` and `port_monitor`, I'd prefer that the examples illustrate that. So no need for one enable/disable with each, I'd be interested in how we'd be using them together.

README.md
```diff
@@ -278,6 +278,7 @@ end
 Performs actions on logical interconnect and associated interconnects.
 
 Note: By default it performs the action `:none`.
+Note 2: In :update_port_monitor action, if the data is informed, this will always precede the port_monitor property.
```
fgbulsoni: Suggestion for clarity:
Note 2: In the `:update_port_monitor` action, if the same field is informed inside both `data` and `port_monitor`, the value of that field inside `data`  will supersede the value inside `port_monitor`.

README.md
```diff
@@ -278,7 +278,7 @@ end
 Performs actions on logical interconnect and associated interconnects.
 
 Note: By default it performs the action `:none`.
-Note 2: In :update_port_monitor action, if the data is informed, this will always precede the port_monitor property.
+Note 2: In the `:update_port_monitor` action, if the same field is informed inside both `data` and `port_monitor`, the value of that field inside data will supersede the value inside `port_monitor`.
```
fgbulsoni: (...) inside `data` will supersede (...)

Since all fields are being marked with `` makes sense to follow the standard.

fgbulsoni (approved):

PR #330, opened 2017-09-20
Importing certificate to power device IPDU if auto_import_cert is true

ricardogpsf:
### Description
Importing certificate to power device IPDU if auto_import_cert property is set to true.
When we are applying discover action to oneview_power_device, some appliances (like appliance of OV 3.10) require the valid certificate to access IPDU server, this commit changes are inserting the code to import the certificate;

### Environment Details
 - **OneView Version:** 3.10
 - **platform:** Any

### Check List
- [ ] Changes documented in the [CHANGELOG](https://github.com/HewlettPackard/oneview-chef/blob/master/CHANGELOG.md).

### DIFF
--- a/README.md
+++ b/README.md
@@ -687,7 +687,7 @@ oneview_power_device '<iPDU hostname>' do
   client <my_client>
   username <username>
   password <password>
-  auto_import_cert [true, false] # Optional. Only used with the :discover action. Default value is false.
+  auto_import_certificate [true, false] # Optional. Only used with the :discover action. Default value is true.
   action :discover
 end
 ```
--- a/README.md
+++ b/README.md
@@ -687,6 +687,7 @@ oneview_power_device '<iPDU hostname>' do
   client <my_client>
   username <username>
   password <password>
+  auto_import_cert [true, false] # Optional. Only used with the :discover action. Default value is false.
   action :discover
 end
 ```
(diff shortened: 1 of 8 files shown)


README.md
```diff
@@ -687,6 +687,7 @@ oneview_power_device '<iPDU hostname>' do
   client <my_client>
   username <username>
   password <password>
+  auto_import_cert [true, false] # Optional. Only used with the :discover action. Default value is false.
```
fgbulsoni: I guess that if we're really trying to make things easier to the end user here, `true` would be the better default.

aalexmonteiro: Suggestion: Maybe it was good that the name was ```import_certificate```. It is more readable.

fgbulsoni (changes requested):

README.md
fgbulsoni: I do like `import_certificate`, but I feel it is not clear that it is doing so automatically. 

In the case of `import_certificate` I'd expect a field to pass in the certificate.
Perhaps `auto_import_certificate` ? 

My naming sense sucks though, If you guys don't feel like it lacks clarity `import_certificate` is also fine by me.

ricardogpsf: I was thinking in client that are using "old" appliances, that don't need import some certificate, they will have to do changes into theirs recipes of power_device, right?! Is this the better way?

ricardogpsf: I'll change the name to `auto_import_certificate`, right?! =D \o/

libraries/resource_providers/api200/power_device_provider.rb
```diff
@@ -23,6 +23,8 @@ def discover
             pd_klass.discover(@item.client, hostname: @name, username: @new_resource.username, password: @new_resource.password)
           rescue OneviewSDK::OneViewError => error
             raise error unless @new_resource.auto_import_certificate && error.message.include?('Unable to retrieve the input certificate')
+            Chef::Log.warn("Oneview returned the following error:\n: #{error.message}")
+            Chef::Log.info("Trying to import certificate for #{@resource_name} '#{@name}'")
```
fgbulsoni: Trying to **automatically** import (...)

fgbulsoni (approved):

PR #334, opened 2017-09-29
API 500 Support for Volume

aalexmonteiro:
### Description
Adding Volumes for HPE OneView API500 support.

### Issues Resolved
N/A

### Check List


libraries/resource_providers/api500/c7000/volume_provider.rb
```diff
@@ -0,0 +1,107 @@
...
+            @item.delete(flag)
+          end
+          true
+        end
+
+        def set_properties
+          @item['properties'] ||= {}
+          @item['isPermanent'] ||= @new_resource.is_permanent
+          @item['properties']['name'] = @new_resource.name
+          @item['properties']['description'] ||= @item['description'] if @item['description']
```
fgbulsoni: Is this `if` necessary?

README.md
```diff
@@ -579,7 +581,12 @@ oneview_volume 'Volume_1' do
   storage_pool <storage_pool_name>
   volume_template <volume_template_name>
   snapshot_pool <snapshot_pool_name>
-  action [:create, :create_if_missing, :delete]
+  properties <volume_properties>                # Hash. Used in create_from_snapshot action only. Only available on API500 and onwards.
+  device_volume_name <device_volume_name>       # String. Name of the external volume that want to add. Used in add action only. Only available on API500 and onwards.
+  is_shareable <is_shareable>                   # Optional <TrueClass, FalseClass> but required in create_from_snapshot. Only available on API500 and onwards.
+  is_permanent <is_permanent>                   # <TrueClass, FalseClass>. Default: 'true'. Only available on API500 and onwards.
+  delete_only_appliance <delete_only_appliance> # <TrueClass, FalseClass>. Default: 'false'. Used in delete action only. If true remove from appliance and storage system, if false remove from appliance only. Only available on API500 and onwards.
```
fgbulsoni: So, this is named delete_**only**_appliance, and if it is `true`, it will remove from appliance **and** storage system?

fgbulsoni (changes requested):

libraries/resource_providers/api200/volume_provider.rb
```diff
@@ -39,21 +39,18 @@ def load_storage_system
           credentials: { ip_hostname: @new_resource.storage_system },
           name: @new_resource.storage_system
         }
-        @item.set_storage_system(load_resource(:StorageSystem, data))
+        @storage_system = resource_named(:StorageSystem).new(@item.client, data)
+        @item.set_storage_system(@storage_system)
       end
 
       def load_storage_pool
```
fgbulsoni: This whole method is literally one line long:
```ruby
@item['storagePoolUri'] = load_resource(:StoragePool, { name: @new_resource.storage_pool, storageSystemUri: @storage_system['uri'] }, :uri)
```
Is there a need for it?

libraries/resource_providers/api200/volume_provider.rb
```diff
@@ -39,21 +39,18 @@ def load_storage_system
...
-        raise 'Must specify a storage_system to use the storage_pool helper.' unless @item['storageSystemUri']
-        sp = resource_named(:StoragePool).find_by(@item.client, name: @new_resource.storage_pool, storageSystemUri: @item['storageSystemUri']).first
-        raise "Storage Pool '#{sp['name']}' not found" unless sp
-        @item['storagePoolUri'] = sp['uri']
+        data = { name: @new_resource.storage_pool, storageSystemUri: @storage_system['uri'] }
+        @item['storagePoolUri'] = load_resource(:StoragePool, data, 'uri')
       end
 
       def load_snapshot_pool
```
fgbulsoni: Same here
```ruby
@item['snapshotPoolUri'] = load_resource(:StoragePool, { name: @new_resource.snapshot_pool, storageSystemUri: @storage_system['uri'] }, :uri)
```

libraries/resource_providers/api500/c7000/volume_provider.rb
aalexmonteiro: This is not really necessary, just in ```snapshotPool(only for storeServ)``` and ```dataProtectionLevel(only for storeVirtual)```.

README.md
```diff
@@ -579,7 +581,12 @@ oneview_volume 'Volume_1' do
   storage_pool <storage_pool_name>
   volume_template <volume_template_name>
   snapshot_pool <snapshot_pool_name>
-  action [:create, :create_if_missing, :delete]
+  properties <volume_properties>                # Hash. Used in create_from_snapshot action only. Only available on API500 and onwards.
+  device_volume_name <device_volume_name>       # String. Name of the external volume that want to add. Used in add action only. Only available on API500 and onwards.
```
ricardogpsf: Why not to use the name passed to resource? 
Like here: 
```ruby
  oneview_volume 'Volume_1'` do
    ...
  end
```
We can use the name "Volume_1" to the action `add`, right? And we would avoid a new property.

libraries/resource_providers/api500/c7000/volume_provider.rb
```diff
@@ -0,0 +1,107 @@
...
+        def load_storage_pool
+          load_storage_system
+          super
+        end
+
+        def load_volume_template
+          template = resource_named(:VolumeTemplate).new(@item.client, name: @new_resource.volume_template)
+          @item.set_storage_volume_template(template)
+          @item['storagePoolUri'] = template['storagePoolUri']
+          @item['snapshotPoolUri'] = template['storagePoolUri'] if template['family'] == 'StoreServ'
```
ricardogpsf: Must item[**'snapshotPoolUri'**] receive the value from template[**'storagePoolUri'**] ? Or would be ` @item['snapshotPoolUri'] = template['snapshotPoolUri']`?

libraries/resource_providers/api500/c7000/volume_provider.rb
```diff
@@ -0,0 +1,107 @@
...
+        end
+
+        def create_from_snapshot
+          validate_required_properties(:properties)
+          raise("#{@resource_name} '#{@name}' not found!") unless @item.retrieve!
+          properties = convert_keys(@new_resource.properties, :to_s)
+          properties['isShareable'] ||= @new_resource.is_shareable
+          return Chef::Log.info "Volume '#{properties['name']}' already exists" if resource_named(:Volume).new(@item.client, name: properties['name']).retrieve!
```
ricardogpsf: Why not use `load_resource` instead of `resource_named(:Volume).new(@item.client, name: properties['name']).retrieve!`? 
It will be code more clean and already throw error when the resource is not found.

ricardogpsf (changes requested):

README.md
aalexmonteiro: Because the ```device_volume_name``` and volume ```name``` are not the same thing. I can import a volume named ```X``` and want this volume named ```Y```.

libraries/resource_providers/api500/c7000/volume_provider.rb
aalexmonteiro: I can't to raise an exception here. I only return a chef::log if volume already exists.

ricardogpsf: Ah, sorry! I saw wrong, you are returning the Log and not raising exception. 😅

libraries/resource_providers/api500/c7000/volume_provider.rb
aalexmonteiro: The template doesn't have this attribute, but there is one ```template['properties']['snapshotPool']['default']``` , I will use it. This attribute has the snapshotUri assigned in creating the template volume.

README.md
fgbulsoni: "# String. Name of external volume to be added. Used only in the `add` action. Only available on API500 and onwards."

fgbulsoni: "# String. Name of external volume to be added. Used only in `add` action. Available on API500 and onwards."

libraries/matchers.rb
```diff
@@ -60,7 +60,7 @@
     oneview_unmanaged_device:           %i[add remove add_if_missing],
     oneview_uplink_set:                 standard_actions,
     oneview_user:                       standard_actions,
-    oneview_volume:                     standard_actions + %i[create_snapshot delete_snapshot],
+    oneview_volume:                     standard_actions + %i[add create_from_snapshot create_snapshot delete_snapshot],
```
fgbulsoni: Why are we using `add` now? With this, we'd have both `create` and `add` for this resource, and I don't see the particular need for this. Especially since in API500 we really do create the volumes anyway.
Any reasoning behind it?

libraries/resource_providers/api500/c7000/volume_provider.rb
```diff
@@ -0,0 +1,103 @@
...
+        end
+
+        def add_or_edit
+          validate_required_properties(:storage_system, :device_volume_name, :is_shareable)
+          options = Marshal.load(Marshal.dump(@item.data))
+          return Chef::Log.info "#{@resource_name} '#{@name}' already exists" if @item.exists?
+          storage_system = resource_named(:StorageSystem).new(@item.client, name: @new_resource.storage_system)
+          Chef::Log.info "Adding #{@resource_name} '#{@name}'"
+          @context.converge_by "Added #{@resource_name} '#{@name}'" do
+            resource_named(:Volume).add(@item.client, storage_system, @new_resource.device_volume_name, @new_resource.is_shareable, options)
```
fgbulsoni: Gotta say I don't particularly like this `add` statement, since it differs so much from the usual `add(client, data)`. 

I understand this is implemented this way in the SDK at the moment, but it would probably be better for standardization purposes to keep it in the usual format, and extract the info from data. Or have helpers to help build the final object. 

It just feels off to me, out of the expected for the `add` action to expect these many info and not get a good body for a post directly.

fgbulsoni (changes requested):

libraries/resource_providers/api200/volume_provider.rb
```diff
@@ -18,11 +18,12 @@ def load_resource_with_associated_resources
...
           @item.set_storage_volume_template(resource_named(:VolumeTemplate).new(@item.client, name: @new_resource.volume_template))
         else # Can't set the storage_pool or snapshot_pool if we specify a volume_template
-          load_storage_system if @new_resource.storage_system
+          validate_required_properties(:storage_system, :storage_pool)
+          storage_system_uri = load_storage_system
           # @item.set_storage_pool(resource_named(:StoragePool).new(@item.client, name: storage_pool)) if storage_pool
           # Workaround for issue in oneview-sdk:
-          load_storage_pool if @new_resource.storage_pool
-          load_snapshot_pool if @new_resource.snapshot_pool
+          @item['storagePoolUri'] = load_resource(:StoragePool, { name: @new_resource.storage_pool, storageSystemUri: storage_system_uri }, 'uri')
```
fgbulsoni: Just a reminder that since this does not have the `if @new_resource.storage_pool` like it did previously, it *will* fail and raise and error if @new_resource.storage_pool is *nil*. Might be the wanted behavior, but just pointing out in case it isn't.

libraries/resource_providers/api200/volume_provider.rb
```diff
@@ -18,11 +18,12 @@ def load_resource_with_associated_resources
         if @new_resource.volume_template
           @item.set_storage_volume_template(resource_named(:VolumeTemplate).new(@item.client, name: @new_resource.volume_template))
         else # Can't set the storage_pool or snapshot_pool if we specify a volume_template
-          load_storage_system if @new_resource.storage_system
+          validate_required_properties(:storage_system, :storage_pool)
+          storage_system_uri = load_storage_system
           # @item.set_storage_pool(resource_named(:StoragePool).new(@item.client, name: storage_pool)) if storage_pool
           # Workaround for issue in oneview-sdk:
```
fgbulsoni: Is this issue still around? I'd likely add in the number of the issue in the comment if so.

libraries/resource_providers/api200/volume_provider.rb
```diff
@@ -18,11 +18,12 @@ def load_resource_with_associated_resources
         if @new_resource.volume_template
           @item.set_storage_volume_template(resource_named(:VolumeTemplate).new(@item.client, name: @new_resource.volume_template))
         else # Can't set the storage_pool or snapshot_pool if we specify a volume_template
-          load_storage_system if @new_resource.storage_system
+          validate_required_properties(:storage_system, :storage_pool)
+          storage_system_uri = load_storage_system
```
fgbulsoni: This method is called exactly 1 time and does not offer a lot.
It is likely better to drop the method and just write its contents here directly.

```ruby
storage_system = resource_named(:StorageSystem).new(@item.client, {credentials: { ip_hostname: @new_resource.storage_system }, name: @new_resource.storage_system})
@item.set_storage_system(storage_system)
```
Then use `storage_system['uri']` instead of `storage_system_uri` on the following calls.

fgbulsoni (changes requested):

libraries/matchers.rb
aalexmonteiro: In API 500, the endpoints are different as well as the payload. For this reason, I believe that the two must be separated, because the two actions are different. In the add action, I don't create a volume, I add an existing volume, and the documentation does not cite how I check if that volume is an external volume, and that I can understand that it is an addition and not a creation. Perhaps the way to differentiate one or the other is by passing the device_volume_name property. What do you think?

README.md
```diff
@@ -579,7 +581,12 @@ oneview_volume 'Volume_1' do
...
+  is_shareable <is_shareable>                   # Optional <TrueClass, FalseClass> but required in create_from_snapshot. Only available on API500 and onwards.
+  is_permanent <is_permanent>                   # <TrueClass, FalseClass>. Default: 'true'. Only available on API500 and onwards.
+  delete_only_appliance <delete_only_appliance> # <TrueClass, FalseClass>. Default: 'false'. Used in delete action only. If true remove from appliance only, if false remove from appliance and storage system. Only available on API500 and onwards.
```
tmiotto: I recommend changing the name to `delete_from_appliance_only`

libraries/resource_providers/api200/volume_provider.rb
```diff
@@ -18,11 +18,12 @@ def load_resource_with_associated_resources
...
           # @item.set_storage_pool(resource_named(:StoragePool).new(@item.client, name: storage_pool)) if storage_pool
```
tmiotto: I guess this was already fixed in Ruby SDK, check if this comment is still valid.

resources/volume.rb
```diff
@@ -38,3 +43,15 @@
 action :delete_snapshot do
   OneviewCookbook::Helper.do_resource_action(self, :Volume, :delete_snapshot)
 end
+
+action :create_from_snapshot do
+  OneviewCookbook::Helper.do_resource_action(self, :Volume, :create_from_snapshot)
+end
+
+action :create_from_snapshot do
```
tmiotto: Why is there two `:create_from_snapshot` actions

spec/unit/resources/volume/create_by_template_spec.rb
```diff
@@ -4,14 +4,32 @@
   let(:resource_name) { 'volume' }
   include_context 'chef context'
 
-  before :each do
-    allow_any_instance_of(OneviewSDK::Volume).to receive(:exists?).and_return(false)
+  let(:base_sdk) { OneviewSDK }
```
tmiotto: `OneviewSDK::API200`

spec/unit/resources/volume/delete_spec.rb
```diff
@@ -4,15 +4,55 @@
   let(:resource_name) { 'volume' }
   include_context 'chef context'
 
+  let(:target_class) { OneviewSDK::Volume }
```
tmiotto: `OneviewSDK::API200::Volume`

spec/unit/resources/volume/create_spec.rb
```diff
@@ -4,27 +4,44 @@
...
-      ]
-    )
-    allow_any_instance_of(OneviewSDK::StorageSystem).to receive(:exists?).and_return(false)
-    allow_any_instance_of(OneviewCookbook::API200::VolumeProvider).to receive(:load_resource).and_call_original
-    allow_any_instance_of(OneviewCookbook::API200::VolumeProvider)
-      .to receive(:load_resource)
-      .with(:StorageSystem, anything)
-      .and_return(OneviewSDK::StorageSystem.new(client, name: 'StorageSystem1', uri: '/rest/storage-systems/1'))
-    allow_any_instance_of(OneviewSDK::Volume).to receive(:exists?).and_return(false)
+  let(:base_sdk) { OneviewSDK }
```
tmiotto: `OneviewSDK::API200`

tmiotto (changes requested):

libraries/resource_providers/api500/c7000/volume_provider.rb
```diff
@@ -57,14 +53,14 @@ def create_from_snapshot
           end
         end
 
-        def add_or_edit
-          validate_required_properties(:storage_system, :device_volume_name, :is_shareable)
-          options = Marshal.load(Marshal.dump(@item.data))
-          return Chef::Log.info "#{@resource_name} '#{@name}' already exists" if @item.exists?
-          storage_system = resource_named(:StorageSystem).new(@item.client, name: @new_resource.storage_system)
+        def add_if_missing
```
aalexmonteiro: I think add_if_missing more coherent, since editing can be done through create action. If you don't agree, I will change to the previous form.

libraries/resource_providers/api500/c7000/volume_provider.rb
```diff
@@ -20,33 +20,29 @@ def load_resource_with_associated_resources
           else
             validate_required_properties(:storage_system, :storage_pool)
             storage_system_uri = load_storage_system
-            @item['storagePoolUri'] = load_resource(:StoragePool, { name: @new_resource.storage_pool, storageSystemUri: storage_system_uri }, 'uri')
-            @item['snapshotPoolUri'] = load_resource(:StoragePool, { name: @new_resource.snapshot_pool, storageSystemUri: storage_system_uri }, 'uri') if @new_resource.snapshot_pool
+            @item.set_storage_pool(resource_named(:StoragePool).new(@item.client, name: @new_resource.storage_pool, storageSystemUri: storage_system_uri))
+            @item.set_snapshot_pool(resource_named(:StoragePool).new(@item.client, name: @new_resource.snapshot_pool, storageSystemUri: storage_system_uri)) if @new_resource.snapshot_pool
           end
-          set_properties unless @item.exists?
+          @item['properties']['name'] = @new_resource.name
```
aalexmonteiro: As ```set_storage_pool``` and ```set_snapshot_snapshot``` are called, the ```properties``` attribute is present in the data because in the API500, the ```name``` property is inside the ```properties```. In the ```exists?``` method in the SDK will search by name within ```properties```. If you do not think this is cool, I'll have to improve the methods in the SDK as describe [here](https://github.com/HewlettPackard/oneview-sdk-ruby/issues/282).

libraries/resource_providers/api500/c7000/volume_provider.rb
```diff
@@ -20,33 +20,29 @@ def load_resource_with_associated_resources
...
+          @item.exists? ? @item.data.delete('properties') : set_properties
         end
 
         def load_storage_system
-          data = {
-            hostname: @new_resource.storage_system,
-            name: @new_resource.storage_system
-          }
-          load_resource(:StorageSystem, data, 'uri')
+          load_resource(:StorageSystem, { hostname: @new_resource.storage_system, name: @new_resource.storage_system }, 'uri')
```
aalexmonteiro: I did not drop this method as requested on provider 200 because I call it in more than one location, so I thought it would be best to keep it for a single point of maintenance.

tmiotto (approved):

ricardogpsf (approved):
=D

libraries/resource_providers/api500/c7000/volume_provider.rb
```diff
@@ -0,0 +1,94 @@
...
+            validate_required_properties(:storage_system, :storage_pool)
+            storage_system_uri = load_storage_system
+            @item.set_storage_pool(resource_named(:StoragePool).new(@item.client, name: @new_resource.storage_pool, storageSystemUri: storage_system_uri))
+            @item.set_snapshot_pool(resource_named(:StoragePool).new(@item.client, name: @new_resource.snapshot_pool, storageSystemUri: storage_system_uri)) if @new_resource.snapshot_pool
+          end
+          @item['properties']['name'] = @new_resource.name
+          @item.exists? ? @item.data.delete('properties') : set_properties
+        end
+
+        def load_storage_system
```
fgbulsoni: Any point to having this one-liner method?

aalexmonteiro: I did not drop this method as requested on provider 200 because I call it in more than one location, so I thought it would be best to keep it for a single point of maintenance

fgbulsoni: Yeah, but we end up calling the method in two places, instead of that one line of code in two places. Not really a difference worth having a method in this case.

As mentioned in conversation, if this specific load were to be changed in future versions, there *is* a definite advantage to having it in a method, since then we'd just need to overwrite this method instead of the ones calling it. 

However, since that does not appear to be the case at the moment, at least that we know of... It is not adding anything by being a separate method, so I'd suggest dropping it.

fgbulsoni (approved):

PR #335, opened 2017-09-29
Removing new_profile action from oneview_server_profile_template

### LINKED ISSUE #333: Wrong behavior of new_profile action of oneview_server_profile_template when using API500
### Scenario/Intent
When we are using API500 to apply action `:new_profile` of `oneview_server_profile_template`, the resource is trying create a new Server Profile Template based on `server_profile_name` property, but the Server Profile is not created yet, throwing the exception. Even that Server Profile is created, the action `:new_profile` should create a new Server Profile and not a Server Profile Template, because the action `:new_profile` is about create a new Server Profile based on the template.
After exchange ideas with @fgbulsoni, he raised an idea of to remove the action `:new_profile`, and I think that is good idea, because with the ability to create recipes of oneview_server_profile associating it to a Server Profile Template, the `:new_profile` action make less sense now.

### Expected Result
Remove the `:new_profile` action of oneview_server_profile_template resource OR fix the bug.
I votes to remove the `:new_profile` action. What do you think about it? =)

### Actual Result
When ServerProfileTemplateProvider API200 call the [new_profile](https://github.com/HewlettPackard/oneview-chef/blob/cd610a3c7b0fe708aa3081916e262aacd35caa08/libraries/resource_providers/api200/server_profile_template_provider.rb#L39-L44) method to create a new ServerProfile, it execute a processing and it calls the `create_if_missing ` method that calls the [create](https://github.com/HewlettPackard/oneview-chef/blob/cd610a3c7b0fe708aa3081916e262aacd35caa08/libraries/resource_providers/api500/c7000/server_profile_template_provider.rb#L18-L26) method of ServerProfileTemplateProvider of API500 is called, that try `load_resource` ServerProfile and throw a error because the ServerProfile was not found (it was not created yet).
This behavior is wrong, the `new_profile` method should just create a new Server Profile.

ricardogpsf:
### Issues Resolved
#333 consequently was fixed with this;
#336 

### Check List


aalexmonteiro (approved):

fgbulsoni (commented):
This is fine by me as I've mentioned before, I don't really see any advantage in managing a `Server Profile` in a `Server Profile Template` resource when the same could be done in the `Server Profile` resource.
But, since this will cause a breaking change, I'm leaving the final vote for @tmiotto .

CHANGELOG.md
```diff
@@ -52,6 +52,9 @@ Bug fixes:
 - [#243](https://github.com/HewlettPackard/oneview-chef/issues/243) Chef-client 13.2.20 does not allow modification of property :save_resource_info
 - [#180](https://github.com/HewlettPackard/oneview-chef/issues/180) Create Mixins for the resource providers common methods
 
+### Breaking changes
+- The `new_profile` action of oneview_server_profile_template was removed.
```
tmiotto: I suggest saying that it was "permanently replaced by" the corresponding action.

tmiotto (changes requested):
Raise also the version to 3.0.0

CHANGELOG.md
ricardogpsf: ok, but that action was not replaced in fact. 
What do you think if we use this below text?
"The `:new_profile` action of oneview_server_profile_template was permanently removed. Instead, use the `:create` action of server_profile setting the `server_profile_template` property."

fgbulsoni: Perhaps something like: 
The `:new_profile` action was permanently replaced by specifying a `server_profile_template` on creates and updates of the `server_profile` resource.

fgbulsoni: Perhaps better:
The `:new_profile` action was incorporated into create and update actions of the `server_profile` resource where a `server_profile_template` is specified.

tmiotto (approved):

tmiotto (changes requested):
Change README.MD

ricardogpsf:
Oh, Good! Thanks!

README.md
```diff
@@ -777,9 +777,9 @@ oneview_server_profile_template 'ServerProfileTemplate1' do
   ethernet_network_connections <ethernet_network_connections_data>
   fc_network_connections <fc_network_connections_data>
   network_set_connections <network_set_connections_data>
-  server_profile_name <profile_name>  # String - Optional. In :new_profile action represents the name of the Server Profile to created. In create/create_if_missing is the name of the Server Profile be used as base for the Server Profile Template. Only available on API500 and onwards to be used as base.
+  server_profile_name <profile_name>  # String - Optional. Used in create/create_if_missing actions. It is the name of the Server Profile be used as base for the Server Profile Template. Only available on API500 and onwards to be used as base.
```
fgbulsoni: Used in create and create_if_missing

tmiotto (commented):
Is there an issue opened for this change? If it is, add it to the changelog.

ricardogpsf:
@tmiotto The issue #333  is about the bug, but in this case we are doing a enhancemente, really need add that issue?

tmiotto:
It would be food to have this change logged into the issues. You can create another one to remove the method, and close the #333 referring to the new one, and then add it to the changelog

tmiotto (approved):

PR #350, opened 2018-03-01
Update to string instead of integer

jjasghar:
- The readme and examples say integer when it is supposed to be a
  string. This PR changes the examples and the README to reflect it.

/cc @russellseymour

Signed-off-by: JJ Asghar <jj@chef.io>

### DIFF
--- a/README.md
+++ b/README.md
@@ -1016,7 +1016,7 @@ end
   my_client = { url: 'https://oneview.example.com', user: 'Administrator', password: 'secret123' }
 
   eth_net_data = {
-    vlanId: 50,
+    vlanId: "50",
     purpose: 'General',
     smartLink: false,
     privateNetwork: false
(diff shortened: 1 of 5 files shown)


madhav-bharadwaj:
Hi @jjasghar , Can you please provide us references where it says `vlanId` needs to be a string? 
Thanks.

jjasghar:
We got an error when running the code. We converted it to a string and it worked.

madhav-bharadwaj:
Can you Please share the steps to reproduce this error? This attribute is an integer and not a string according to product documentation. 
We will reproduce this and check for the root cause.

russellseymour:
After investigation to reproduce this error and provide you with a stack trace, I have found that the recipe will compile if the `vlanId` is set as an integer.

I have been unable to reproduce the inital error I saw several times so this is a non-issue.

jjasghar:
Per @russellseymour 's comment I'll close this.

